PR #25, opened 2015-01-20
Call fetchResultsDataSourceHasResults: delegate method with NO

### LINKED ISSUE #21: `fetchResultsDataSourceHasResults:` isn't called if there are actually no results
Expected behavior: 
`[delegate fetchResultsDataSourceHasResults:NO]` should be called by the data sources

brockboland:
When there are no results in the fetched results controller, the number of sections is 0, so it doesn't bother calling `collectionView:numberOfItemsInSection:` or `tableView:numberOfRowsInSection:`, as the case may be, so the delegate `fetchResultsDataSourceHasResults:` method was never called with a NO argument.

This change calls that delegate method in the numberOfSections method if the section count is 0.

Fixes #21

### DIFF
--- a/Pod/Classes/Optional Data Sources/VOKCollectionDataSource.m
+++ b/Pod/Classes/Optional Data Sources/VOKCollectionDataSource.m
@@ -116,7 +116,7 @@
 {
     NSInteger sectionCount = [[self.fetchedResultsController sections] count];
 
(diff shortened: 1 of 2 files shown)


seanwolter:
yayayayayayaya
the 3 year old bug is dead

seanwolter:
Good to pull

PR #32, opened 2015-02-23
Handle fetch limit

brockboland:
Handles #29 and #30. Since `NSFetchedResultsController` ignores the fetchLimit, we need to handle it ourselves.

@vokal/ios-developers code review please?

Also: how should we test this? We don't have KIF or FRY in this project right now, and it might be a good place to try out FRY. Any objections?

PR #33, opened 2015-02-25
Respect fetchLimit when updating items, not just sections

brockboland:
I messed up in #32: our `controllerDidChangeContent:` method handles section and item changes separately…for…reasons? I had implemented and tested #32 in a separate project, then copied the code over piece-by-piece (because that project had an older version of Vokoder, this was easier than replacing the entire file and doing the `git add -p` dance), and missed that this logic is basically repeated twice in that method.

@vokal/ios-developers code review please? Especially from someone more familiar with the collection view data source?


seanwolter:
LGTM

seanwolter:
aaand `pod trunk push`ed

PR #36, opened 2015-04-09
Import UIKit in the fetched results data source

brockboland:
I guess I've never tried to use this without importing UIKit from the Prefix header or something. Or, maybe something changed in Xcode 6.3, but on a fresh project, Xcode was griping that it doesn't know anything about UITableViewDataSource or delegate.

@vokal/ios-developers I'm surprised no one else has run into this, which leads me to believe that I'm doing it wrong. What do you think?

### DIFF
--- a/Pod/Classes/Optional Data Sources/VOKFetchedResultsDataSource.h
+++ b/Pod/Classes/Optional Data Sources/VOKFetchedResultsDataSource.h
@@ -3,6 +3,8 @@
 //  CoreData
 //
 
+#import <UIKit/UIKit.h>
+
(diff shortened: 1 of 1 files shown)


chillpop:
LGTM

seanwolter:
Odd. I just created a new project with Xcode 6.3 and I can't find a pch

seanwolter:
Good to pull.

brockboland:
Would you be so kind as to tag 1.1.5?

seanwolter:
I think you need to update the podspec before I tag it.

brockboland:
Forgot about that

PR #42, opened 2015-05-13
Set NSFRC's delegate to nil on dealloc

MattFoley:
What do you guys think of this? I was finding that intermittently NSFRC would send messages to a deallocated datasource, as the delegate wasn't set to nil by anything else. Seemed to only happen when lots of data and view churn was happening simultaneously.

### DIFF
--- a/Pod/Classes/Optional Data Sources/VOKFetchedResultsDataSource.m
+++ b/Pod/Classes/Optional Data Sources/VOKFetchedResultsDataSource.m
@@ -407,10 +407,7 @@
          self.tableView.dataSource = nil;
     }
 
-    if (_fetchedResultsController.delegate) {
-        _fetchedResultsController.delegate = nil;
-    }
-    
+    _fetchedResultsController.delegate = nil;
 }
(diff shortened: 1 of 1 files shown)


seanwolter:
Oh hell yeah. Open source!

Pod/Classes/Optional Data Sources/VOKFetchedResultsDataSource.m
```diff
@@ -406,6 +406,10 @@ - (void)dealloc
     if (self.tableView.dataSource == self) {
          self.tableView.dataSource = nil;
     }
+
+    if (_fetchedResultsController.delegate) {
```
seanwolter: Why check before setting to `nil`? If it's already `nil` no harm done.

MattFoley: That's true, lemme update it.

seanwolter:
Travis is an idiot.
Good to pull!

PR #49, opened 2015-09-15
Add an assertion to the mapper

vokal-isaac:
This is to help in future debugging of an issue I just helped someone debug.

### DIFF
--- a/Pod/Classes/VOKManagedObjectMapper.m
+++ b/Pod/Classes/VOKManagedObjectMapper.m
@@ -190,6 +190,11 @@
 
         Class managedObjectClass = objc_getClass(className);
         objc_property_t prop = class_getProperty(managedObjectClass, propertyName);
+        if ([self isKindOfClass:[VOKManagedObjectDefaultMapper class]]) {
+            NSAssert(prop, @"Property named %s doesn't exist on class named %s.  Did you mean to use the default mapper?", propertyName, className);
+        } else {
+            NSAssert(prop, @"Property named %s doesn't exist on class named %s.", propertyName, className);
+        }
 
         NSString *attributeString = [NSString stringWithCString:property_getAttributes(prop) encoding:NSUTF8StringEncoding];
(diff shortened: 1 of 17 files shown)


designatednerd:
( :goat: , :goat: , :goat: )

designatednerd:
probably want to bump the podspec too though

vokal-isaac:
Maybe?  I wasn't sure if this change was worth pushing out a new version.

designatednerd:
probably a new patch version

Pod/Classes/VOKManagedObjectMapper.m
```diff
@@ -190,6 +190,11 @@ - (Class)expectedClassForObject:(NSManagedObject *)object andKey:(id)key
 
         Class managedObjectClass = objc_getClass(className);
         objc_property_t prop = class_getProperty(managedObjectClass, propertyName);
+        if ([self isKindOfClass:[VOKManagedObjectDefaultMapper class]]) {
```
chillpop: If this is run in a release environment with assertions turned off, will this empty if statement be optimized out by the compiler?  If not, does it make sense to refactor this to make sure it doesn't do this check in release builds?

vokal-isaac: Since the `NSAssert()` statements become no-ops, I'd expect the optimization step to take the conditional out entirely, since it'd branch to nothing in both cases.

chillpop: :+1:

vokal-isaac:
I've bumped the podspec version to 1.3.2.  Should I `pod update Vokoder` in the sample projects, too?

designatednerd:
yes plz. If you can add a test that it throws when you're expecting it to, that'd be helpful.

vokal-isaac:
Pod update and additional tests pushed.

designatednerd:
( :goat: , :goat: , :goat: ), 2.0. @chillpop, anything further?

chillpop:
LGTM

designatednerd:
Tagged + `pod trunk push`ed

PR #52, opened 2015-10-20
Custom bundles

designatednerd:
Add the ability to pass in a custom bundle, mostly to facilitate use with Swift.

### DIFF
--- a/Pod/Classes/VOKCoreDataManager.h
+++ b/Pod/Classes/VOKCoreDataManager.h
@@ -68,7 +68,7 @@ typedef void(^VOKObjectIDsReturnBlock)(NSArray *arrayOfManagedObjectIDs);
  The main context will be initialized immediately. This method should only be called from the main queue.
  @param resource    The filename of the mom or momd file in your project. If nil the first model found in the provided bundle will be used.
  @param database    The filename of the SQLite store in your application. A nil database name will create an in-memory store.
- @param bundle      The bundle where the Managed Object Model can be found. 
+ @param bundle      The bundle where the Managed Object Model can be found. If nil, the bundle for this class will be used. 
  */
 - (void)setResource:(NSString *)resource
            database:(NSString *)database
--- a/Pod/Classes/VOKCoreDataManager.h
+++ b/Pod/Classes/VOKCoreDataManager.h
@@ -68,7 +68,7 @@ typedef void(^VOKObjectIDsReturnBlock)(NSArray *arrayOfManagedObjectIDs);
  The main context will be initialized immediately. This method should only be called from the main queue.
  @param resource    The filename of the mom or momd file in your project. If nil the first model found in the provided bundle will be used.
  @param database    The filename of the SQLite store in your application. A nil database name will create an in-memory store.
- @param bundle      The bundle where the S
+ @param bundle      The bundle where the Managed Object Model can be found. 
  */
 - (void)setResource:(NSString *)resource
            database:(NSString *)database
--- a/Pod/Classes/VOKCoreDataManager.h
+++ b/Pod/Classes/VOKCoreDataManager.h
@@ -56,7 +56,7 @@ typedef void(^VOKObjectIDsReturnBlock)(NSArray *arrayOfManagedObjectIDs);
 /**
  Set the name of the managed object model and the name of the SQL lite store on disk. Call this first when you setup the core data stack.
  The main context will be initialized immediately. This method should only be called from the main queue.
- NOTE: This assumes your managed object file is in the same bundle as this class. If it isn't (perhaps because you're using CocoaPods and Swift), use setResource:database:bundle: instead.
+ NOTE: This assumes your managed object file is in the main bundle. If it isn't, use setResource:database:bundle: instead.
  @param resource    The filename of the mom or momd file in your project. If nil the first model found in this class's bundle will be used.
  @param database    The filename of the SQLite store in your application. A nil database name will create an in-memory store.
  */
@@ -68,7 +68,7 @@ typedef void(^VOKObjectIDsReturnBlock)(NSArray *arrayOfManagedObjectIDs);
  The main context will be initialized immediately. This method should only be called from the main queue.
  @param resource    The filename of the mom or momd file in your project. If nil the first model found in the provided bundle will be used.
  @param database    The filename of the SQLite store in your application. A nil database name will create an in-memory store.
- @param bundle      The bundle where the Managed Object Model can be found. If nil, the bundle for this class will be used. 
+ @param bundle      The bundle where the Managed Object Model can be found. If nil, the main bundle will be used. 
  */
 - (void)setResource:(NSString *)resource
            database:(NSString *)database
--- a/Pod/Classes/VOKCoreDataManager.h
+++ b/Pod/Classes/VOKCoreDataManager.h
@@ -54,13 +54,25 @@ typedef void(^VOKObjectIDsReturnBlock)(NSArray *arrayOfManagedObjectIDs);
 - (NSManagedObjectModel *)managedObjectModel;
 
 /**
- Set the name of the managed object model and the name of the SQL lite store on disk. Call this first when you setup the core data stack. 
+ Set the name of the managed object model and the name of the SQL lite store on disk. Call this first when you setup the core data stack.
  The main context will be initialized immediately. This method should only be called from the main queue.
- @param resource    The filename of the mom or momd file in your project. If nil the first model found in the main bundle will be used.
+ NOTE: This assumes your managed object file is in the same bundle as this class. If it isn't (perhaps because you're using CocoaPods and Swift), use setResource:database:bundle: instead.
+ @param resource    The filename of the mom or momd file in your project. If nil the first model found in this class's bundle will be used.
  @param database    The filename of the SQLite store in your application. A nil database name will create an in-memory store.
  */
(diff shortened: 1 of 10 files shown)


designatednerd:
@vokal/ios-developers Who wants to take a look?

brockboland:
on it

Pod/Classes/VOKCoreDataManager.h
```diff
@@ -54,15 +54,27 @@ typedef void(^VOKObjectIDsReturnBlock)(NSArray *arrayOfManagedObjectIDs);
...
  */
 - (void)setResource:(NSString *)resource
            database:(NSString *)database;
 
 /**
+ Set the name of the managed object model and the name of the SQL lite store on disk, in the provided bundle. Call this first when you setup the core data stack.
+ The main context will be initialized immediately. This method should only be called from the main queue.
+ @param resource    The filename of the mom or momd file in your project. If nil the first model found in the provided bundle will be used.
+ @param database    The filename of the SQLite store in your application. A nil database name will create an in-memory store.
+ @param bundle      The bundle where the S
```
brockboland: Bundle where the…?

designatednerd: where the sentence fragment lives, duh.

Fixin'

brockboland:
LGTM! Anyone else want to have a look-see?

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -112,13 +127,23 @@ - (NSPersistentStoreCoordinator *)persistentStoreCoordinator
...
+
 #pragma mark - Initializers
 
 - (void)initManagedObjectModel
 {
-    NSURL *modelURL = [[NSBundle bundleForClass:[self class]] URLForResource:self.resource withExtension:@"momd"];
+    NSURL *modelURL = [self.bundleForModel URLForResource:self.resource withExtension:@"momd"];
     if (!modelURL) {
-        modelURL = [[NSBundle bundleForClass:[self class]] URLForResource:self.resource withExtension:@"mom"];
+        modelURL = [self.bundleForModel URLForResource:self.resource withExtension:@"mom"];
```
JilScott: i<3Mom

JilScott:
LGTM

Pod/Classes/VOKCoreDataManager.h
```diff
@@ -54,15 +54,27 @@ typedef void(^VOKObjectIDsReturnBlock)(NSArray *arrayOfManagedObjectIDs);
...
  */
 - (void)setResource:(NSString *)resource
            database:(NSString *)database;
 
 /**
+ Set the name of the managed object model and the name of the SQL lite store on disk, in the provided bundle. Call this first when you setup the core data stack.
+ The main context will be initialized immediately. This method should only be called from the main queue.
+ @param resource    The filename of the mom or momd file in your project. If nil the first model found in the provided bundle will be used.
+ @param database    The filename of the SQLite store in your application. A nil database name will create an in-memory store.
+ @param bundle      The bundle where the Managed Object Model can be found.
```
chillpop: maybe add "If nil, the bundle for this class will be used"?

designatednerd: :+1:

SampleProject/VOKCoreDataManagerTests/BundleTests.m
```diff
@@ -0,0 +1,30 @@
...
+@implementation BundleTests
+
+- (void)testWorksWithoutSettingCustomBundle
+{
+    XCTAssertNoThrow([[VOKCoreDataManager sharedInstance] setResource:@"VICoreDataModel" database:nil]);
+}
+
+- (void)testWorksWithSettingCustomBundle
+{
+    XCTAssertNoThrow([[VOKCoreDataManager sharedInstance] setResource:@"VICoreDataModel"
```
chillpop: how about a test that does throw an exception?  (pass in the test bundle?)

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -112,13 +127,23 @@ - (NSPersistentStoreCoordinator *)persistentStoreCoordinator
...
 }
 
+- (NSBundle *)bundleForModel
+{
+    if (!_bundleForModel) {
+        //Default to using the bundle for this class.
+        _bundleForModel = [NSBundle bundleForClass:[self class]];
+    }
+    
+    return _bundleForModel;
```
vokal-isaac: I'd suggest that you `return _bundleForModel ?: [NSBundle bundleForClass:[self class]];` instead, so that the property is null-resettable.

Also, is using the bundle for the pod a better default than using the main bundle?  (I really don't know.)

designatednerd: How is what you're suggesting going to make that easier to tell? 

I suspect it may not work quite right if we set it up to use the main bundle - it's also much more backwards compatible with what we've already got to leave it as is.

vokal-isaac: Not easier... not easy by any means.  But if `_bundleForModel` is `nil`, then `- bundleForModel` will return the default value, whereas if `_bundleForNil` is non-`nil`, it'll return whatever its value is.  Whereas your code will lose the information that it was intended to be the default value.  :tomato: :tomato: (It was more of a thing when I was thinking about it being null-resettable, but that's already the case with your code, so :shrugs:.)

You're right that it's more backward-compatible in terms of not changing behavior of existing code and that's probably enough reason to leave it alone.  I can't see why making it the main bundle would be a problem, but :shrugs:.

SampleProject/VOKCoreDataManagerTests/BundleTests.m
designatednerd: Hm, ok, well, this is obnoxious - in Xcode 7 `XCTAssertThrows` doesn't actually verify exceptions thrown by `NSAssert` statements and `XCTAssertNoThrow` doesn't actually flunk something that throws an exception - it flunks almost coincidentally *because* of the exception.

This is probably related to the fact that a `try/catch` block doesn't actually catch the exception thrown by the `NSAssert`, because I tried that too. :stuck_out_tongue_closed_eyes: 

@vokal-isaac @chillpop any thoughts on how to better test this?

chillpop: test that managedObjectModel isn't nil?

designatednerd: That won't actually work since the test will crash on the assertion failure before it gets there :cry:

designatednerd: I still don't have the faintest fucking clue why this wasn't working earlier, but it's working now.

NOTE: I did have to update the bundle it was using to the main bundle so that the MOM didn't have to be embedded in the test bundle. @vokal-isaac in theory this shouldn't break anything major, but in practice....should I change this to a major release?

vokal-isaac: Unless there's a reason not to do a major bump, it's probably safer.

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -594,18 +619,22 @@ - (NSURL *)applicationLibraryDirectory
...
+        for (NSPersistentStore *store in stores) {
+            [[self persistentStoreCoordinator] removePersistentStore:store error:nil];
+            if (self.databaseFilename) {
+                [[NSFileManager defaultManager] removeItemAtPath:store.URL.path error:nil];            
+            }
         }
-    }
+        
+        _persistentStoreCoordinator = nil;
+    } //else the persistent store coordinator is already nil and a bunch of this will crash.
```
vokal-isaac: This shouldn't have crashed when `self.persistentStoreCoordinator` was `nil`...

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -594,18 +619,22 @@ - (NSURL *)applicationLibraryDirectory
 
 - (void)resetCoreData
 {
-    NSArray *stores = [[self persistentStoreCoordinator] persistentStores];
+    if (_persistentStoreCoordinator) {
```
vokal-isaac: Is there a reason for the direct ivar access?

designatednerd: Because the method actually tries to spin up the whole stack again if nothing's there, and if you were testing, say, a bundle which doesn't have a managed object model to make sure it throws, then when it gets spun up again here then it will throw again.

Pod/Classes/VOKCoreDataManager.m
designatednerd: (see above)

Pod/Classes/VOKCoreDataManager.m
vokal-isaac: Ahhh!  In that case, changing the deleted line 597 above to
```objc
NSArray *stores = [_persistentStoreCoordinator persistentStores];
```
should be sufficient to fix it (without needing the rest of the changes).  I'd also suggest adding a comment explaining why the direct ivar access is necessary.

designatednerd:
@vokal-isaac Changes pushed.

vokal-isaac:
Unless you want to make the major version bump, LGTM.

designatednerd:
I don't *think* it's necessary. 

I guess I'll find out if it breaks the shit out of everything.

PR #54, opened 2015-10-23
Address #34.

### LINKED ISSUE #34: Guard against NULL issue in default mapper
It's in or around
```objc
        Class managedObjectClass = objc_getClass(className);
        objc_property_t prop = class_getProperty(managedObjectClass, propertyName);

        NSString *attributeString = [NSString stringWithCString:property_getAttributes(prop) encoding:NSUTF8StringEncoding];
```
in `expectedClassForObject:andKey:`.  (I think it's that `prop` is `NULL` and then `property_getAttributes(prop)` crashes.)

vokal-isaac:
Resolves #34.

Avoid production crashes with additional safety around failure to get a property and/or failure to get a class.  (Programmer-error cases are already covered by assertions.)

Since @chillpop said he was going to version-bump in a subsequent PR, I didn't touch the version and didn't `pod update` in either sample project.

@vokal/ios-developers Code review please?

### DIFF
--- a/Pod/Classes/VOKManagedObjectMapper.m
+++ b/Pod/Classes/VOKManagedObjectMapper.m
@@ -156,6 +156,10 @@
 - (id)checkClass:(id)inputObject managedObject:(NSManagedObject *)object key:(NSString *)key
 {
     Class expectedClass = [self expectedClassForObject:object andKey:key];
+    if (!expectedClass) {
+        VOK_CDLog(@"Failed to get class for %@\nProperty: %@", object, key);
+        return nil;
+    }
     if (![inputObject isKindOfClass:expectedClass]) {
         if (!(self.ignoreOptionalNullValues
               && (!inputObject || [[NSNull null] isEqual:inputObject])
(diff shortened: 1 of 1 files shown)


designatednerd:
For context: #34

Pod/Classes/VOKManagedObjectMapper.m
```diff
@@ -156,6 +156,10 @@ - (id)checkString:(id)outputObject withDateFormatter:(NSDateFormatter *)dateForm
 - (id)checkClass:(id)inputObject managedObject:(NSManagedObject *)object key:(NSString *)key
 {
     Class expectedClass = [self expectedClassForObject:object andKey:key];
+    if (!expectedClass) {
+        VOK_CDLog(@"Failed to get class for %@\nProperty: %@", object, key);
+        return nil;
```
designatednerd: Should this also return `Nil`? And should we explicitly be returning a `Class` rather than an `id` if that's the case?

vokal-isaac: `checkClass:managedObject:key:` returns `inputObject` on success and `nil` on failure, so `id` and `nil` are correct.

designatednerd: in that case, should it be `nil` instead of `Nil` below?

vokal-isaac: Nope, that's in `- (Class)expectedClassForObject:(NSManagedObject *)object andKey:(id)key`

designatednerd: ahhhhh derp, right. Thank you!

designatednerd:
( :goat: , :goat: , :goat: )

chillpop:
LGTM

PR #55, opened 2015-10-23
Nullability annotations

chillpop:
- Added nullability annotations to Vokoder headers.
- updated Travis to use Xcode 7.1

Tested on both Xcode 7.1 and 6.4

@vokal/ios-developers, code review?

As before, I have still more changes to come.  I'm waiting to bump the version number and run pod update on the example projects until there are more changes.


brockboland:
on it

vokal-isaac:
Would it make sense to include something like
```objc
#if !__has_feature(nullability)
#define NS_ASSUME_NONNULL_BEGIN
#define NS_ASSUME_NONNULL_END
// … some other #defines so that the nullability things that don't exist are defined to nothing and won't cause compiler errors
#endif
```
so that this doesn't become unusable on pre-Xcode-6.4?

Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.h
```diff
@@ -70,7 +73,8 @@ typedef void(^VOKManagedObjectsReturnBlock)(NSArray *arrayOfManagedObjects);
  @param contextOfNil    The managed object context to update and/or insert the object. If nil, the main context will be used.
  @return                An instance of this subclass of NSManagedObject.
  **/
-+ (instancetype)vok_addWithDictionary:(NSDictionary *)inputDict forManagedObjectContext:(NSManagedObjectContext *)contextOrNil;
++ (nullable instancetype)vok_addWithDictionary:(nullable NSDictionary *)inputDict
```
designatednerd: Should the dictionary actually be nullable? If you're trying to import a nonexistent dictionary, that's just gonna fail anyway.

designatednerd:
@vokal-isaac Do we have anyone still using 6.4?

chillpop:
nullability was added in 6.3

designatednerd:
@chillpop Didn't they change the way it works in 7 though?

chillpop:
in 7, they changed the names to _Nullable and _Nonnull but __nullable and __nonnull are mapped to those in 7.

brockboland:
Am I the only one that kind find details on the difference between `__Nullable` and `nullable` and `__nullable`?

vokal-isaac:
@designatednerd I don't know if anyone's actively using 6.anything.  Maybe ask on Slack or check with lead(s) on any projects that started before Xcode 7 was released?  (If I had an in-flight tight-timeline project that started on 6.3 or 6.4, I'd keep it there until the timeline was met, unless something forced me off...)

chillpop:
@brockboland (and others) this explains the details a bit: https://developer.apple.com/swift/blog/?id=25

chillpop:
ELS is still on 6.4 but this will work back to 6.3.

chillpop:
It's not too much of a change to support Xcode versions without nullability but it will make the property declarations uglier and possibly harder to scan for the nullability:
current
`@property (nullable, weak, readonly) UITableView *tableView;`
revised:
`@property (weak, readonly) UITableView * __nullable tableView;`
(with `#define __nullable` to make it a no-op)
I"m not sure it's worth supporting older versions of Xcode that we don't seem to have many projects (if any) actively using.

vokal-isaac:
I'm not sure I particularly disagree about supporting older versions... but if we're going to break compatibility, I'd prefer to package up all the non-breaking changes into a patch or minor release, then put the breaking changes in a major release.

seanwolter:
Wouldn't older Xcode users specify an older version of Vokoder? Oh. Isaac beat me to it. Great work!
:ghost: open-source ghost out! :ghost:

chillpop:
In the interests of keeping this moving, I did what Isaac suggested.  Updates pushed.

chillpop:
Additional updates pushed.  I realized that some headers don't include the main header so I had to put the degradation of nullability qualifiers in its own file.  This required a version bump and pod update on the sample projects to get them to include the new file.

designatednerd:
Agreeing with @vokal-isaac that this should probably be packaged up with #58 for a major release, so you'll need to bump the podspec to that. But otherwise: LGTM.

chillpop:
I've bumped the podspec to 1.4.0.  This should still be backwards compatible.

designatednerd:
@chillpop: What version of Xcode is this backwards compatible to? We should probably note that in the README.

chillpop:
Just checked: it builds and runs in Xcode 5.1

designatednerd:
Wow. I'm impressed that you still HAVE Xcode 5.1. 

In that case: LGTM. @vokal-isaac @brockboland and oh why not @seanwolter anyone else have thoughts?

chillpop:
I was surprised that I still had it too.

chillpop:
updated README pushed.  Xcode 5 runs the example, but the tests use async testing methods that weren't introduced until Xcode 6.

designatednerd:
SLGTM

chillpop:
I'll wait until noon and then merge this.

PR #57, opened 2015-10-25
Copyright and Macro

designatednerd:
Things which can be updated without a major version bump.


seanwolter:
lgtm

chillpop:
Given that this is under an MIT license, I don't think "All rights reserved" is correct.  I don't know what _is_ correct though.  IANAL (BIAMTO)

chillpop:
perhaps including the license text in each file?

designatednerd:
@chillpop IANAL either, but if I recall correctly one of the rights we reserve is to open-source it. Maybe update it to point to the `LICENSE.md` file for licensing info? I'd rather not copy-pasta that text into every file.

chillpop:
maybe "Some rights reserved" or nothing after the copyright?  "All rights reserved" should mean you can't do anything with it which is a contradiction to the license file.

Additionally, I think the license file should have the license name on the first line like so: http://choosealicense.com/licenses/mit

designatednerd:
I always prefer https://tldrlegal.com/license/mit-license since it's reasonably simple.

designatednerd:
I'll switch it to nothing after the copyright and add a link to the LICENSE.

designatednerd:
@chillpop Changes pushed.

LICENSE
```diff
@@ -1,4 +1,6 @@
-Copyright (c) 2014-2015 Vokal 
+https://tldrlegal.com/license/mit-license
+
+Copyright (c) 2014-2015 Vokal
```
chillpop: looking through this diff, we have some code going back to 2012...

chillpop:
LGTM, aside from changing to 2012-2015

designatednerd:
@chillpop Fixed.

chillpop:
:+1:

LICENSE
```diff
@@ -1,4 +1,6 @@
-Copyright (c) 2014-2015 Vokal 
+https://tldrlegal.com/license/mit-license
```
vokal-isaac: If you look at the full text there, they have "The MIT License (MIT)" on the first line...

Pod/Classes/Internal/VOKCoreDataManagerInternalMacros.h
```diff
@@ -3,7 +3,7 @@
 //  SampleProject
 //
 //  Created by Isaac Greenspan on 5/21/14.
-//
+//  Copyright © 2015 Vokal.
```
vokal-isaac: 2014–2015?

Pod/Classes/Optional Data Sources/VOKCollectionDataSource.h
```diff
@@ -2,7 +2,7 @@
 //  VICollectionDataSource.h
 //
 //  Created by teejay on 5/6/13.
-//  Copyright (c) 2013 Vokal. All rights reserved.
+//  Copyright © 2013 Vokal.
```
vokal-isaac: 2013–2015?

etc.

Pod/Classes/VOKManagedObjectMap.h
```diff
@@ -32,6 +34,16 @@
...
 
+/**
+ *  Creates a map for a string to a number using a default NSNumberFormatter.
+ *
+ *  @param inputKeyPath             The foreign key to match with the local key.
+ *  @param coreDataSelectorSymbol   The local selector symbol
+ */
+#ifndef VOK_MAP_FOREIGN_STRING_TO_LOCAL_NUMBER
+#   define VOK_MAP_FOREIGN_STRING_TO_LOCAL_NUMBER(inputKeyPath, coreDataSelectorSymbol) [VOKManagedObjectMap mapWithForeignKeyPath:inputKeyPath coreDataKey:VOK_CDSELECTOR(coreDataSelectorSymbol) numberFormatter:[NSNumberFormatter new]]
+#endif
```
vokal-isaac: To me, it would be preferable for `VOKManagedObjectMap` to have a default number formatter that was used when the types aligned and when no number formatter were set.  I think this could be achieved by overriding the getter of the `numberFormatter` property to return `_numberFormatter ?: [self vok_defaultNumberFormatter]` and creating `- (NSNumberFormatter *)vok_defaultNumberFormatter` to return a `dispatch_once`-initialized number formatter...

SampleProject/Models/VIThing.m
```diff
@@ -3,7 +3,7 @@
 //  SampleProject
 //
 //  Created by Sean Wolter on 8/23/14.
-//
+//  Copyright © 2015 Vokal.
 //
 
 #import "VIThing.h"
```
vokal-isaac: For the major-version-bump release, would it be reasonable to refactor all the `VI` classes to be `VOK` classes?

designatednerd: Probably, but not here.

Pod/Classes/Optional Data Sources/VOKCollectionDataSource.h
designatednerd: I think that should be reasonably covered by the license.

Pod/Classes/VOKManagedObjectMap.h
designatednerd: I don't know that having a `dispatch_once` numberformatter would work since you could have numbers in different formats (decimals, money, integers, etc).

vokal-isaac: It'd work just as well as what you've got here, which is just using a stock number formatter.  The code here, though, will create a new stock number formatter each time it's used.

designatednerd: I think I'm confused here - wouldn't this mean that the same number formatter would be used for every single object? Or are you saying this would be a per-object dispatch_once?

designatednerd:
(merge conflicts fixed, btw)

Pod/Classes/VOKManagedObjectMap.h
vokal-isaac: Yes, for every single object that uses this macro or, alternately, uses the proposed default behavior.  As you have it now, they're all using different ones that all behave exactly the same way.

chillpop:
@vokal-isaac, @designatednerd, I've updated this PR.

designatednerd:
LGTM :+1:

Pod/Classes/VOKManagedObjectMap.h
```diff
@@ -35,6 +37,16 @@
...
 
+/**
+ *  Creates a map for a string to a number using a default NSNumberFormatter.
+ *
+ *  @param inputKeyPath             The foreign key to match with the local key.
+ *  @param coreDataSelectorSymbol   The local selector symbol
+ */
+#ifndef VOK_MAP_FOREIGN_STRING_TO_LOCAL_NUMBER
+#   define VOK_MAP_FOREIGN_STRING_TO_LOCAL_NUMBER(inputKeyPath, coreDataSelectorSymbol) [VOKManagedObjectMap mapWithForeignKeyPath:inputKeyPath coreDataKey:VOK_CDSELECTOR(coreDataSelectorSymbol) numberFormatter:[VOKManagedObjectMap vok_defaultNumberFormatter]]
+#endif
```
vokal-isaac: My preference is still that, rather than adding this macro, we fold this behavior into the inherent mapper behavior.  That is, if the foreign value is a string and the local value is a number and there's no number formatter set, use the default number formatter.

chillpop: I would prefer to be explicit about mapping numbers.  What if you have a "number-like" string that you want to stay a string? (e.g.: primary key identifier?)

chillpop: actually, never mind. per our discussion, I'll rip this out and make a new PR with those changes.

chillpop:
update pushed

vokal-isaac:
Per our discussion, LGTM.

PR #58, opened 2015-10-25
Unique key, drop iCarousel support, fix VI -> VOK

designatednerd:
NOTE: These changes are dependent on #57 and should not be merged until that one gets merged down. They also should not be merged until we're ready to release a new major version since they change a couple major things about how the mapping works.


Pod/Classes/VOKCoreDataManager.m
```diff
@@ -93,7 +95,7 @@ - (NSManagedObjectContext *)tempManagedObjectContext
         return nil;
     }
 
-    NSManagedObjectContext *tempManagedObjectContext = [[NSManagedObjectContext alloc] initWithConcurrencyType:NSConfinementConcurrencyType];
+    NSManagedObjectContext *tempManagedObjectContext = [[NSManagedObjectContext alloc] initWithConcurrencyType:NSPrivateQueueConcurrencyType];
```
seanwolter: I thought `NSPrivateQueueConcurrencyType` required the use of `performBlock:` (and related methods) as the context creates and manages its own private queue. To me, `NSConfinementConcurrencyType` meant we managed our queues manually.

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -524,6 +583,36 @@ - (NSManagedObjectContext *)temporaryContext
...
 }
 
+- (NSManagedObjectContext *)temporaryInMemoryContext
+{
+    //Spin up a new context
+    NSManagedObjectContext *context;
+    context = [[NSManagedObjectContext alloc] initWithConcurrencyType:NSPrivateQueueConcurrencyType];
+    
+    //Use the same MOM as the main context.
+    NSManagedObjectContext *mainContext = [[VOKCoreDataManager sharedInstance] managedObjectContext];
```
seanwolter: I think this will barf if called off the main queue if the main context doesn't already exist since `initManagedObjectContext` 
Couldn't you get self.managedObjectModel directly in the line below?

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -296,6 +323,38 @@ - (NSArray *)importArray:(NSArray *)inputArray forClass:(Class)objectClass withC
     return [returnArray copy];
 }
 
+- (NSArray *)arrayToImportAfterImportingRepeatedUniqueKeys:(NSArray *)repeatedUniqueKeys
+                                   forForeignUniqueKeyPath:(NSString *)foreignUniqueKeyPath
+                                           usingInputArray:(NSArray *)inputArray
+                                                  forClass:(Class)objectClass
+                                               withContext:(NSManagedObjectContext *)contextOrNil
+{
+    NSMutableArray *normalNumberOfKeys = [NSMutableArray arrayWithArray:inputArray];
```
chillpop: I think this variable could have a better name, it's not an NSNumber.  Maybe `uniquedInputArray`?

SampleProject/VOKCoreDataManagerTests/ManagedObjectImportExportTests.m
```diff
@@ -597,6 +599,34 @@ - (void)testUniqueKeyPath
     XCTAssertEqualObjects([[self customDateFormatter] stringFromDate:person.birthDay], dateString);
 }
 
+- (void)testImportingTwoObjectsWithTheSameUniqueIDOverwritesWithUniqueKey
```
chillpop: is it worth adding a test to check that any import after this will trigger an assert? [line 310](https://github.com/designatednerd/Vokoder/blob/major_version_changes/Pod/Classes/VOKCoreDataManager.m#L310)

chillpop:
that's all from me.

Pod/Classes/VOKCoreDataManager.m
chillpop: `NSConfinementConcurrencyType` is deprecated as of iOS 9.  We'll need to change to something else.  Maybe we should allow passing in the concurrency type and default to `NSMainQueueConcurrencyType`?

chillpop: In any case, this change should be documented in the README or the header comments.

seanwolter: I suppose the correct course is to follow Apple's lead and start using the context's queues rather than managing our own.

Pod/Classes/VOKCoreDataManager.m
designatednerd: good point.

SampleProject/VOKCoreDataManagerTests/ManagedObjectImportExportTests.m
designatednerd: In theory any import after this would just update the existing object.

chillpop: then maybe add a test to make sure it _doesn't_ throw an exception?

Pod/Classes/VOKCoreDataManager.m
designatednerd: Probably. I'll have to do a bit more research on this before I can officially figure out this bit.

Pod/Classes/Optional Data Sources/VOKCarouselDataSource.m
```diff
@@ -1,183 +0,0 @@
...
-    //remember to always set any properties of your carousel item
-    //views outside of the `if (view == nil) {...}` check otherwise
-    //you'll get weird issues with carousel item content appearing
-    //in the wrong place in the carousel
-    label.text = [(self.fetchedObjects)[index] stringValue];
-    
-    return view;
-}
-
-@end
```
brockboland: byeeee

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -256,11 +258,36 @@ - (NSArray *)importArray:(NSArray *)inputArray forClass:(Class)objectClass withC
...
     if (mapper.uniqueComparisonKey) {
-        NSArray *arrayOfUniqueKeys = [inputArray valueForKeyPath:mapper.foreignUniqueComparisonKey];
+        NSString *foreignUniqueKey = mapper.foreignUniqueComparisonKey;
+        NSArray *arrayOfUniqueKeys = [inputArray valueForKeyPath:foreignUniqueKey];
         //filter out all NSNull's
         arrayOfUniqueKeys = [arrayOfUniqueKeys filteredArrayUsingPredicate:[NSPredicate predicateWithFormat:@"self != nil"]];
+        
+        //Use a counted set to grab any repeated keys and notify the user
+        NSCountedSet *countedKeySet = [[NSCountedSet alloc] initWithArray:arrayOfUniqueKeys];
+        for (id key in countedKeySet) {
```
brockboland: Suggestion: wrap this whole thing (the for loop and conditional below) in a check to see if it's even necessary:

```objc
if (countedKeySet.count < arrayOfUniqueKeys.count)
```

From the NSCountedSet docs:

>  The count method defined by the superclass NSSet has special significance; it returns the number of distinct objects, not the total number of times objects are represented in the set.

designatednerd: ooh, good call.

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -524,6 +583,35 @@ - (NSManagedObjectContext *)temporaryContext
...
+    //Spin up a new context
+    NSManagedObjectContext *context;
+    context = [[NSManagedObjectContext alloc] initWithConcurrencyType:NSPrivateQueueConcurrencyType];
+    
+    //Use the same MOM as the main context.
+    NSManagedObjectModel *model = self.managedObjectModel;
+    
+    //Use an in-memory store
+    NSPersistentStoreCoordinator *coordinator;
+    coordinator = [[NSPersistentStoreCoordinator alloc] initWithManagedObjectModel:model];
```
brockboland: This is personal preference, but I'm not a fan of using two lines to define and then set variables like this (and `context`)

designatednerd: Oh, yeah. I just copy-pasta'd the code from the thing, you're correct that this looks bizarre.

chillpop:
Travis failed: you missed some VI -> VOK changes:
https://travis-ci.org/vokal/Vokoder/jobs/87688440#L273-L291

designatednerd:
@chillpop DAMN YOU OTHER SAMPLE PROJECT I FORGOT EXISTED. This is now fixed.

chillpop:
LGTM, still waiting for #57?

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -296,6 +327,38 @@ - (NSArray *)importArray:(NSArray *)inputArray forClass:(Class)objectClass withC
...
+                                   forForeignUniqueKeyPath:(NSString *)foreignUniqueKeyPath
+                                           usingInputArray:(NSArray *)inputArray
+                                                  forClass:(Class)objectClass
+                                               withContext:(NSManagedObjectContext *)contextOrNil
+{
+    NSMutableArray *uniquedInputArray = [NSMutableArray arrayWithArray:inputArray];
+    
+    NSMutableArray *itemsToRemove = [NSMutableArray array];
+    for (id repeatedKey in repeatedUniqueKeys) {
+        for (id item in inputArray) {
```
vokal-isaac: If the goal is to import these in order, should these two `for` loops be flipped?

This and the code above using the counted set make me uncomfortable, but I can't put my finger on why.  Are there tests added to cover all the corners of the intended behavior?

vokal-isaac: Actually, could this be:
```objc
for (id item in inputArray) {
    if ([repeatedUniqueKeys containsObject:[item valueForKeyPath:foreignUniqueKeyPath]]) {
```
(including replacing the `if` statement right below here).

designatednerd: I tried flipping them but it winds up causing problems  - I can't remember off the top of my head what they were but it definitely did not work.

designatednerd:
@chillpop I still need to look into the nonsense with the private queue concurrency type before this gets merged down

Pod/Classes/VOKCoreDataManager.m
vokal-isaac: Ohh... I think I see why and now the comment on line 352–3 makes some sense.  Hrm.

vokal-isaac: I'd like to take a shot at fiddling with this code, preferable pre-merge, if that's okay.

chillpop:
Does the concurrency queue issues need to happen in this current set of PRs?

chillpop:
(I say that partly because I have more changes in the wings, waiting for these to be merged)

Pod/Classes/VOKCoreDataManager.m
vokal-isaac: ... I'm confused and not entirely sure I understand what the intended behavior is with duplicated unique keys.  Anyone want to / have time to explain it to me?

designatednerd: Intended behavior is 1) Warn the user that there's duplicated unique keys in the same import 2) Import those items sequentially so if you have multiple items with the same unique key, they just overwrite in the order in which you received them.

chillpop:
I removed the tempInMemoryContext and reverted the concurrency type change on the temp context (deferring whatever will change there until later).  Waiting on Isaac to add some comments to his version of the unique key situation.

vokal-isaac:
It'll be a bit before I get to it, but I expect to get to it today.

designatednerd:
@chillpop I merged @vokal-isaac's PR to my branch in.

chillpop:
LGTM

designatednerd:
@chillpop Would you prefer to tackle the confinement concurrency stuff in a separate PR? I can change the name of this PR to better reflect what it actually contains in that case.

chillpop:
Yes, I'd prefer to separate changes to concurrency into a separate PR after some thought.

designatednerd:
OK, name of PR changed to better reflect final contents. LGTM - @chillpop @vokal-isaac?

chillpop:
LGTM

chillpop:
@vokal-isaac, any further comments?

vokal-isaac:
Nope, good on my end.

PR #59, opened 2015-10-26
Objective-C generics with backwards compatibility

chillpop:
Added Obj-C generic decorations to method declarations.  

I `typedef`d named specialized collection types with fallbacks to unspecialized/undecorated types for earlier versions of Xcode.

@vokal/ios-developers, @seanwolter, code review?

For versioning, this can also be a part of 1.4.0 from #55 since it will still compile under older versions of Xcode.

### DIFF
--- a/Paging Data Source Example/Paging Data Source Example.xcodeproj/project.pbxproj
+++ b/Paging Data Source Example/Paging Data Source Example.xcodeproj/project.pbxproj
@@ -378,7 +378,7 @@
 				GCC_PREFIX_HEADER = "Paging Data Source Example-Prefix.pch";
 				INFOPLIST_FILE = "$(SRCROOT)/Paging Data Source Example-Info.plist";
 				IPHONEOS_DEPLOYMENT_TARGET = 7.0;
-				PRODUCT_BUNDLE_IDENTIFIER = "com.VOKAL.${PRODUCT_NAME:rfc1034identifier}";
+				PRODUCT_BUNDLE_IDENTIFIER = "com.Vokal.Paging-Data-Source-Example";
 				PRODUCT_NAME = "Paging Data Source Example";
 				WRAPPER_EXTENSION = app;
 			};
@@ -396,7 +396,7 @@
 				GCC_SYMBOLS_PRIVATE_EXTERN = NO;
 				INFOPLIST_FILE = "$(SRCROOT)/Paging Data Source Example-Info.plist";
 				IPHONEOS_DEPLOYMENT_TARGET = 7.0;
-				PRODUCT_BUNDLE_IDENTIFIER = "com.VOKAL.${PRODUCT_NAME:rfc1034identifier}";
+				PRODUCT_BUNDLE_IDENTIFIER = "com.Vokal.Paging-Data-Source-Example";
 				PRODUCT_NAME = "Paging Data Source Example";
 				WRAPPER_EXTENSION = app;
 			};
--- a/Paging Data Source Example/Paging Data Source Example.xcodeproj/project.pbxproj
+++ b/Paging Data Source Example/Paging Data Source Example.xcodeproj/project.pbxproj
@@ -189,7 +189,7 @@
 			isa = PBXProject;
 			attributes = {
 				CLASSPREFIX = VI;
-				LastUpgradeCheck = 0610;
+				LastUpgradeCheck = 0710;
 			};
 			buildConfigurationList = EA5F199A15C196500089AA44 /* Build configuration list for PBXProject "Paging Data Source Example" */;
 			compatibilityVersion = "Xcode 3.2";
@@ -312,8 +312,10 @@
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				COPY_PHASE_STRIP = NO;
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
+				ENABLE_TESTABILITY = YES;
 				GCC_C_LANGUAGE_STANDARD = "compiler-default";
 				GCC_DYNAMIC_NO_PIC = NO;
+				GCC_NO_COMMON_BLOCKS = YES;
 				GCC_OPTIMIZATION_LEVEL = 0;
 				GCC_PREPROCESSOR_DEFINITIONS = (
 					"DEBUG=1",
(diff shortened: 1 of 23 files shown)


Pod/Classes/VOKCoreDataCollectionTypes.h
```diff
@@ -0,0 +1,61 @@
...
+#define VOKCoreDataCollectionTypes_h
+
+#if __has_feature(objc_generics)
+//forward declare classes used here since we don't know where this will be included
+@class NSManagedObject;
+@class NSManagedObjectID;
+@class VOKManagedObjectMap;
+
+//include the pointer for this one since the fallback is id
+#define VOKManagedObjectSubclassPtr __kindof NSManagedObject *
```
vokal-isaac: What's the `__kindof` do?  Why isn't the fallback `NSManagedObject *`?

chillpop: `__kindof` tells the compiler that subclasses of the type are valid.  The previous method signature where this was used returned `id` before; I kept it the same for the fallback version.  I agree that it should probably be `NSManagedObject *` though. I'll change it.

vokal-isaac: Array and dictionary types like this don't accept subclasses by default?  Is there any other place in ObjC where that's the case?  Is the same true in Swift?

chillpop: I think it's specific to obj-c generics.  Swift lets you do that by default.

vokal-isaac: If I'm reading http://www.miqu.me/blog/2015/06/09/adopting-objectivec-generics/ correctly, `__kindof` allows pulling things out of the collection into subclass types without warnings, but putting subclass type objects into the collection shouldn't require `__kindof`?

Pod/Classes/VOKCoreDataCollectionTypes.h
```diff
@@ -0,0 +1,61 @@
+//
+//  VOKCoreDataCollectionTypes.h
+//  Pods
+//
+//  Created by Carl Hill-Popper on 10/23/15.
+//
+//
+
+#ifndef VOKCoreDataCollectionTypes_h
```
seanwolter: Is `typedef`ing collection types common? I don't know no fancy Swift or modern Objective-C.
it's very cool nonetheless

seanwolter:
niiiiiiice

Paging Data Source Example/Paging Data Source Example.xcodeproj/project.pbxproj
```diff
@@ -392,6 +396,7 @@
 				GCC_SYMBOLS_PRIVATE_EXTERN = NO;
 				INFOPLIST_FILE = "$(SRCROOT)/Paging Data Source Example-Info.plist";
 				IPHONEOS_DEPLOYMENT_TARGET = 7.0;
+				PRODUCT_BUNDLE_IDENTIFIER = "com.VOKAL.${PRODUCT_NAME:rfc1034identifier}";
```
designatednerd: We should probably update this to be LESS SHOUTY

Pod/Classes/VOKCoreDataCollectionTypes.h
chillpop: yeah. it seems like you can put whatever you want into a typed collection.

Pod/Classes/VOKManagedObjectMapper.h
```diff
@@ -6,15 +6,16 @@
...
 
 #import "VOKManagedObjectMap.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
-typedef void(^VOKPostImportBlock)(NSDictionary *inputDict, NSManagedObject *outputObject);
+typedef void(^VOKPostImportBlock)(VOKStringToObjectDictionary *inputDict, NSManagedObject *outputObject);
 
-typedef void(^VOKPostExportBlock)(NSMutableDictionary *outputDict, NSManagedObject *inputObject);
+typedef void(^VOKPostExportBlock)(VOKStringToObjectMutableDictionary *outputDict, NSManagedObject *inputObject);
```
designatednerd: Remind me: Why are we returning this as a mutable dictionary?

designatednerd: Might also help to add documentation to these two completion block types.

vokal-isaac: Not returning, but passing—because the point of the post-export block is to be able to modify the output dictionary, I think.

Pod/Classes/VOKCoreDataCollectionTypes.h
chillpop: Probably not, but it makes being backwards compatible super easy.  If we drop support for Xcode 6, I'd say drop this fallback mechanism too.

Pod/Classes/VOKManagedObjectMapper.h
chillpop: You could easily make a `mutableCopy` if necessary though.

vokal-isaac: Unless the block is changed to return the dictionary, I think making a copy won't affect the export being done when the block is called, though.

Pod/Classes/VOKCoreDataCollectionTypes.h
chillpop: Actually it does seem common for languages with generics.  Maybe they should stay.

chillpop:
updates pushed

designatednerd:
LGTM

chillpop:
anyone else have additional comments on this?

SampleProject/VOKCoreDataManagerTests/ManagedObjectImportExportTests.m
```diff
@@ -452,7 +452,8 @@ - (void)testFetchWithURI
     person = nil;
     [[[VOKCoreDataManager sharedInstance] managedObjectContext] reset];
 
-    VIPerson *personFromURI = [[VOKCoreDataManager sharedInstance] existingObjectAtURI:uri forManagedObjectContext:nil];
+    VIPerson *personFromURI = (VIPerson *)[[VOKCoreDataManager sharedInstance] existingObjectAtURI:uri
```
vokal-isaac: If this still needs the cast, then I don't see the point in having `__kindof`, but I may well be missing the point (I kind of feel like that's likely... pun maybe intended).

chillpop: It needs the cast on Xcode < 7 to avoid a warning.  Without the `__kindof` and with my changes, this method now returns a `NSManagedObject *` instead of an `id`.

vokal-isaac: Actually, looking it over again, I do see the point of the `__kindof` ones.  I don't see the point in returning `VOKManagedObjectSubclass` or having that as a separately-declared type, if `__kindof` only works in collection generics.

vokal-isaac: Ahhhh.... okay, nevermind then.

vokal-isaac:
LGTM

PR #67, opened 2016-01-18
Swift extensions to Vokoder

chillpop:
Added some swift-ier extensions to VOKCoreDataManager and the NSManagedObject category functions. The extensions return typed subclasses instead of `NSManagedObject` to allow for less casting in swift code.  Also, some nullable parameters now have default values for cleaner function calls.

All the swift code is in an optional subspec of the main pod so it is not included by default.

The swift example project has been updated to use the new extensions.

@vokal/ios-developers, @seanwolter, code review?

PR #68, opened 2016-01-19
Require 8.0 for Swift

brockboland:
The podspec update in #67 does not lint:

> -> Vokoder (2.1.0)
>    - ERROR | [iOS] [Vokoder/Swift] swift: Swift support uses dynamic frameworks and is therefore only supported on iOS > 8.

From the [Cocoapods 0.36 release blog post](http://blog.cocoapods.org/Pod-Authors-Guide-to-CocoaPods-Frameworks/): 

> CocoaPods only supports Swift on OS X 10.9 and newer, and iOS 8 and newer.

More details there.

This change just requires 8.0 for the Swift subspec. I wasn't sure if one could do this, but it seems to work and the tests pass.

@vokal/ios-developers is this a terrible idea?


chillpop:
LGTM

designatednerd:
LGTM :v:

brockboland:
Then it shall be done.

PR #69, opened 2016-01-19
Update version to 2.1.1

brockboland:
@designatednerd @chillpop did a dumb and tagged 2.1.0 before linting the podspec. This just fixes the version number. Cool?

### DIFF
--- a/Vokoder.podspec
+++ b/Vokoder.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = "Vokoder"
-  s.version          = "2.1.0"
+  s.version          = "2.1.1"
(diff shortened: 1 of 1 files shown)


designatednerd:
LGTM

chillpop:
would it be easier to remove the tag and then re-tag it? (it can't be done through github, but can be done on the command line)

brockboland:
It would indeed, and I thought about doing that, but becuase this is a public repo I figured it would be better to play it safe.

vokal-isaac:
If the tag is out there in a repo that other people have access to, there's always a risk that someone else already has the tag and you can't make it go away on their end.

It's much safer to just bump the patch version and move on.

chillpop:
fine by me

PR #70, opened 2016-01-28
Fix vok_entityName for the non-MOGenerator case

vokal-isaac:
Actual change is in 75bdca4; 2eec4d7 is running pod install on both sample projects.

Way back in 2014, Vokoder (actually, its predecessor) assumed that entity names were the same as class names.  I tended to name my entities without the project prefix, so that didn't work for me.  In changing it, I leveraged MOGenerator's `entityName` method (or anyone else's, really) if it exists, but otherwise went to look up the entity name in the managed object model.

Except that I did it wrong.  It returned the entity name of the first class that was the same or a superclass of the target class, going through an alphabetical list.  This PR fixes the way entity names are looked up against the model, looking for the entity name for the exact class, then if no match was found, trying the superclass, and going up the superclass chain until it finds an entity or fails.  This'll find the closest ancestor entity name, which is what we want.

I'm applying this to 1.4.0 to make 1.4.1 in this PR.  I'll open a separate PR to apply this to the latest 2.x release, too.

@vokal/ios-developers @seanwolter Code review please?

### DIFF
--- a/Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
+++ b/Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
@@ -51,9 +51,10 @@
             // Start with the class we are...
             Class workingClass = self;
             do {
+                NSString *workingClassName = NSStringFromClass(workingClass);
                 // ... check for a matching entity in the model...
                 for (NSEntityDescription *description in model.entities) {
-                    if ([workingClass class] == [NSClassFromString(description.managedObjectClassName) class]) {
+                    if ([workingClassName isEqualToString:description.managedObjectClassName]) {
                         vok_entityName = description.name;
                         break;
                     }
--- a/Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
+++ b/Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
@@ -59,7 +59,7 @@
                     }
                 }
                 // ... and walk up the superclass chain...
-                workingClass = class_getSuperclass(workingClass);
+                workingClass = [workingClass superclass];
                 // ... until we get Nil or find a matching entity (as long as we have a superclass to test and haven't found the entity name).
             } while (workingClass && !vok_entityName);
         }
--- a/Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
+++ b/Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
@@ -47,12 +47,21 @@
             
             // Since we don't have an entityName class method, look up the entity name in the managed object model.
             NSManagedObjectModel *model = [[VOKCoreDataManager sharedInstance] managedObjectModel];
-            for (NSEntityDescription *description in model.entities) {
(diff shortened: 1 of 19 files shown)


designatednerd:
I'll take a gander

designatednerd:
LGTM

vokal-isaac:
The light is green...

designatednerd:
@vokal/ios-developers Anyone else have thoughts?

Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
```diff
@@ -47,12 +47,21 @@ + (NSString *)vok_entityName
...
-                if ([self isSubclassOfClass:NSClassFromString(description.managedObjectClassName)]) {
-                    vok_entityName = description.name;
-                    break;
+            
+            // Start with the class we are...
+            Class workingClass = self;
+            do {
+                // ... check for a matching entity in the model...
+                for (NSEntityDescription *description in model.entities) {
+                    if ([workingClass class] == [NSClassFromString(description.managedObjectClassName) class]) {
```
chillpop: how about comparing the strings instead of the classes (to avoid calling NSClassFromString for potentially all of the entities repeatedly)?

Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
```diff
@@ -47,12 +47,21 @@ + (NSString *)vok_entityName
...
+                        vok_entityName = description.name;
+                        break;
+                    }
                 }
-            }
+                // ... and walk up the superclass chain...
+                workingClass = class_getSuperclass(workingClass);
```
chillpop: I know you like the runtime, but could you use `[workingClass superclass]`?

Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
vokal-isaac: So like `NSString *workingClassName = NSStringFromClass(workingClass);` at the top of the `do { … }` block and then compare `workingClassName` to `description.managedObjectClassName` using `isEqualToString:`?

chillpop: yes

Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
vokal-isaac: Oops, looks like I should be doing that, yeah.

vokal-isaac:
Changes pushed.

chillpop:
LGTM, how about adding a test or two?

vokal-isaac:
Test pushed.

chillpop:
good tests, good to pull for me

brockboland:
LGTM

designatednerd:
LGTM :three:

designatednerd:
I think this constitutes a quorum. Mergin'

vokal-isaac:
Tag and pod trunk push, please?

PR #72, opened 2016-02-01
Slight refactoring and finally using iOS 5 methods

chillpop:
- use `UIAlertController` if available and eliminate deprecation warning for `UIAlertView`
- eliminate `NSConfinementConcurrencyType` deprecation by using `NSPrivateQueueConcurrencyType` for temp context
- some slight refactoring to combine some methods
- use `performBlock` for managed object context operations (and eliminate the private writing queue)

@vokal/ios-developers, @seanwolter, code review?

### DIFF
--- a/Pod/Classes/VOKCoreDataManager.m
+++ b/Pod/Classes/VOKCoreDataManager.m
@@ -159,7 +159,6 @@ static VOKCoreDataManager *VOK_SharedObject;
             {
                 NSString *title = @"Migration Failed";
                 NSString *message = @"Migration has failed, data will be erased to ensure application stability.";
-#undef __IPHONE_8_0
 #ifdef __IPHONE_8_0 //if compiling with an old version of Xcode that doesn't include the iOS 8 SDK, ignore UIAlertController
                 if ([UIAlertController class]) {
                     UIAlertController *alert = [UIAlertController alertControllerWithTitle:title
--- a/Pod/Classes/VOKCoreDataManager.m
+++ b/Pod/Classes/VOKCoreDataManager.m
@@ -494,7 +494,7 @@ static VOKCoreDataManager *VOK_SharedObject;
     if ([NSOperationQueue mainQueue] == [NSOperationQueue currentQueue]) {
         [self saveContext:self.managedObjectContext];
     } else {
-        [[NSOperationQueue mainQueue] addOperationWithBlock:^{
+        [self.managedObjectContext performBlock:^{
             [self saveContext:self.managedObjectContext];
         }];
     }
@@ -505,7 +505,7 @@ static VOKCoreDataManager *VOK_SharedObject;
     if ([NSOperationQueue mainQueue] == [NSOperationQueue currentQueue]) {
         [self saveContext:self.managedObjectContext];
     } else {
-        [[NSOperationQueue mainQueue] addOperationWithBlock:^{
+        [self.managedObjectContext performBlockAndWait:^{
             [self saveContext:self.managedObjectContext];
         }];
     }
--- a/Pod/Classes/VOKCoreDataManager.m
+++ b/Pod/Classes/VOKCoreDataManager.m
@@ -98,7 +98,12 @@ static VOKCoreDataManager *VOK_SharedObject;
 - (NSManagedObjectModel *)managedObjectModel
 {
     if (!_managedObjectModel) {
-        [self initManagedObjectModel];
+        NSURL *modelURL = [self.bundleForModel URLForResource:self.resource withExtension:@"momd"];
+        if (!modelURL) {
+            modelURL = [self.bundleForModel URLForResource:self.resource withExtension:@"mom"];
+        }
+        NSAssert(modelURL, @"Managed object model not found.");
+        _managedObjectModel = [[NSManagedObjectModel alloc] initWithContentsOfURL:modelURL];
     }
     
     return _managedObjectModel;
@@ -107,7 +112,10 @@ static VOKCoreDataManager *VOK_SharedObject;
 - (NSPersistentStoreCoordinator *)persistentStoreCoordinator
 {
     if (!_persistentStoreCoordinator) {
-        [self initPersistentStoreCoordinator];
+        NSAssert([NSOperationQueue currentQueue] == [NSOperationQueue mainQueue],
+                 @"Must be on the main queue when initializing persistent store coordinator");
+        _persistentStoreCoordinator = [[NSPersistentStoreCoordinator alloc] initWithManagedObjectModel:self.managedObjectModel];
+        [self addPersistentStoreToCoordinator:_persistentStoreCoordinator];
     }
     
     return _persistentStoreCoordinator;
@@ -125,27 +133,12 @@ static VOKCoreDataManager *VOK_SharedObject;
 
 #pragma mark - Initializers
 
-- (void)initManagedObjectModel
+- (void)addPersistentStoreToCoordinator:(NSPersistentStoreCoordinator *)persistentStoreCoordinator
 {
-    NSURL *modelURL = [self.bundleForModel URLForResource:self.resource withExtension:@"momd"];
-    if (!modelURL) {
-        modelURL = [self.bundleForModel URLForResource:self.resource withExtension:@"mom"];
-    }
-    NSAssert(modelURL, @"Managed object model not found.");
-    if (modelURL) {
-        _managedObjectModel = [[NSManagedObjectModel alloc] initWithContentsOfURL:modelURL];
-    }
-}
-
-- (void)initPersistentStoreCoordinator
-{
-    NSAssert([NSOperationQueue currentQueue] == [NSOperationQueue mainQueue], @"Must be on the main queue when initializing persistant store coordinator");
     NSDictionary *options = @{
                               NSMigratePersistentStoresAutomaticallyOption: @YES,
                               NSInferMappingModelAutomaticallyOption: @YES,
                               };
-    
-    _persistentStoreCoordinator = [[NSPersistentStoreCoordinator alloc] initWithManagedObjectModel:self.managedObjectModel];
     
     NSURL *storeURL = [self persistentStoreFileURL];
     NSString *storeType = NSInMemoryStoreType;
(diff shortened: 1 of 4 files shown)


Pod/Classes/VOKCoreDataManager.m
```diff
@@ -12,11 +12,11 @@
 
 #import "VOKMappableModel.h"
 
-@interface VOKCoreDataManager () {
-    NSManagedObjectContext *_managedObjectContext;
-    NSManagedObjectModel *_managedObjectModel;
-    NSPersistentStoreCoordinator *_persistentStoreCoordinator;
```
seanwolter: finally our long national nightmare is over

chillpop: :us: :us: :us:

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -161,8 +145,7 @@ - (void)initPersistentStoreCoordinator
                               NSInferMappingModelAutomaticallyOption: @YES,
                               };
     
-    NSError *error;
-    _persistentStoreCoordinator = [[NSPersistentStoreCoordinator alloc] initWithManagedObjectModel:[self managedObjectModel]];
+    _persistentStoreCoordinator = [[NSPersistentStoreCoordinator alloc] initWithManagedObjectModel:self.managedObjectModel];
```
seanwolter: This may be out of scope for this refactor, buuuut
since `managedObjectModel` and `persistentStoreCoordinator` are now privately read-only properties (instead of raw ivars) would it make sense to change `initPersistentStoreCoordinator` and `initManagedObjectModel` to lazy getters?
If not, I think you can safely use `self.persistentStoreCoordinator` instead of the ivar in the two `init...` methods.

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -171,47 +154,72 @@ - (void)initPersistentStoreCoordinator
...
-            [[NSFileManager defaultManager] removeItemAtPath:storeURL.path error:nil];
-            if (![_persistentStoreCoordinator addPersistentStoreWithType:storeType
-                                                           configuration:nil
-                                                                     URL:storeURL
-                                                                 options:nil
-                                                                   error:&error]) {
-                [NSException raise:@"Vokoder Persistant Store Creation Failure after migration"
-                            format:@"Unresolved error %@, %@", error, [error userInfo]];
+    __block NSError *error;
+    BOOL (^createPersistantStore)() = ^BOOL() {
```
seanwolter: could the block return a BOOL and an NSError?

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -171,47 +154,72 @@ - (void)initPersistentStoreCoordinator
...
+        BOOL success =  [_persistentStoreCoordinator addPersistentStoreWithType:storeType
+                                                                  configuration:nil
+                                                                            URL:storeURL
+                                                                        options:options
+                                                                          error:&localError];
+        error = localError;
+        return success;
+    };
+    
+    if (!createPersistantStore()) {
```
seanwolter: I know I might sound like a dummy, but I think the use of the block here is necessarily complex. Why not simply attempt to create the persistent store, check the bool, and display the rror in the old fashioned way?

SampleProject/VOKAppDelegate.m
```diff
@@ -16,6 +16,10 @@ @implementation VOKAppDelegate
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
 {
+    if (NSClassFromString(@"XCTestCase")) {
```
seanwolter: :sunglasses: so cool

Pod/Classes/VOKCoreDataManager.m
designatednerd: nitpick: persistent store.

Pod/Classes/VOKCoreDataManager.m
chillpop: I think they already function as lazy getters: they are only called from the get methods for `persistentStoreCoordinator` and `managedObjectModel`.  Can you explain more what you would change?

Pod/Classes/VOKCoreDataManager.m
chillpop: Two return values‽  What is this, Swift?

Pod/Classes/VOKCoreDataManager.m
chillpop: I think you're right.  Originally, I made the block because I saw that the two init methods were being called with different params.  Now I see that the difference was inconsequential (no migration options for the second call).

Pod/Classes/VOKCoreDataManager.m
seanwolter: I'd move all the init logic into the getter method. It's code organization I guess. And probably also subjective. I never liked the second init methods.

chillpop: I worry that the getter methods would become bloated and potentially confusing.  How about making the init methods into factory/create methods similar to what I did with `managedObjectContextWithConcurrencyType:`?

seanwolter: :shrug: like I said, it's a matter of opinion
If nobody else dislikes the *sprawling* init methods then I won't mind.

chillpop:
update pushed

seanwolter:
Looking good to pull to me

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -171,47 +147,70 @@ - (void)initPersistentStoreCoordinator
...
+                } else {
+                    //TODO: delete UIAlertView once support is dropped for iOS 7
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+                    [[[UIAlertView alloc] initWithTitle:title
+                                                message:message
+                                               delegate:nil
+                                      cancelButtonTitle:@""
+                                      otherButtonTitles:nil] show];
+#pragma clang diagnostic pop
```
designatednerd: Worth wrapping all this in an `if iOS 7 or lower` block and using UIAlertController otherwise?

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -171,47 +147,70 @@ - (void)initPersistentStoreCoordinator
...
+                //intentional fallthrough
+            case VOKMigrationFailureOptionWipeRecovery:
+                VOK_CDLog(@"Full database delete and rebuild");
+                [[NSFileManager defaultManager] removeItemAtPath:storeURL.path error:nil];
+                if (![persistentStoreCoordinator addPersistentStoreWithType:storeType
+                                                              configuration:nil
+                                                                        URL:storeURL
+                                                                    options:options
+                                                                      error:&error]) {
+                    [NSException raise:@"Vokoder Persistant Store Creation Failure after migration"
```
designatednerd: persistent

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -171,47 +147,70 @@ - (void)initPersistentStoreCoordinator
...
+                                format:@"Unresolved error %@, %@", error, [error userInfo]];
+                }
+                break;
+            case VOKMigrationFailureOptionNone:
+                VOK_CDLog(@"Vokoder Persistant Store Creation Failure: %@", error);
```
designatednerd: persistent

seanwolter: Gee @designatednerd your sure are persistent about spelling persistent correctly. :laughing:  :laughing:  :laughing:

designatednerd: Hey, if @vokal-isaac's not going to do it, someone's got to.

designatednerd: Also you*

seanwolter: :star:

Pod/Classes/VOKCoreDataManager.m
chillpop: something different from `if ([UIAlertController class]) {` on line 162?

chillpop:
spelling updated

Pod/Classes/VOKCoreDataManager.m
chillpop: ...or do you mean a preprocessor statement so this will still compile in Xcode 5 (without iOS 8)?

chillpop:
preprocessor check for iOS 8 pushed

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -171,47 +147,74 @@ - (void)initPersistentStoreCoordinator
...
+                                                            URL:storeURL
+                                                        options:options
+                                                          error:&error]) {
+        switch (self.migrationFailureOptions) {
+            case VOKMigrationFailureOptionWipeRecoveryAndAlert:
+            {
+                NSString *title = @"Migration Failed";
+                NSString *message = @"Migration has failed, data will be erased to ensure application stability.";
+#ifdef __IPHONE_8_0 //if compiling with an old version of Xcode that doesn't include the iOS 8 SDK, ignore UIAlertController
+                if ([UIAlertController class]) {
```
seanwolter: With the compiler check is the class check if/else still required?

chillpop: Yes, otherwise you'd crash when running on iOS < 8.  E.g.: I have an app that I compile on Xcode 6 that is built against the iOS 8 SDK but supports iOS 7 too.

Pod/Classes/VOKCoreDataManager.m
designatednerd: I did mean a preprocessor statement :+1:

designatednerd:
LGTM :+1:

chillpop:
@seanwolter, what say you?

Pod/Classes/VOKCoreDataManager.m
seanwolter: oh yeeeah

seanwolter:
Good to be pulled for me, personally

chillpop:
tagged and pushed to trunk

PR #74, opened 2016-02-08
Use xUnique for example projects

chillpop:
Use xUnique for example projects to make project file diffs easier to handle going forward.

@vokal/ios-developers, @seanwolter "code" review?


designatednerd:
Seems legit to me pending Travis

designatednerd:
@chillpop I think you have to uncheck and recheck the test targets in the test bit of the scheme because the identifiers changed

chillpop:
updated

seanwolter:
I have no opinion about xUnique. Look cool :sunglasses:

vokal-isaac:
@seanwolter You should have an opinion about it and your opinion should be that it's a good thing.  https://github.com/truebit/xUnique

vokal-isaac:
(:goat:,)

PR #73, opened 2016-02-07
New mapping macros using VOKKeyPathHelper

chillpop:
@vokal-isaac started with this:
- New mapping macros using the VOKUtilities/VOKKeyPathHelper for better type checking.
- Deprecate the `VOK_CDSELECTOR` and `VOK_MAP_FOREIGN_TO_LOCAL` macros.

I took over and did this:
- Muck around with module maps to allow using a "non-modular header inside of a framework module"
- Make a new subspec for the mapping macros
- Update Travis to use Xcode 7.2

@vokal/ios-developers, @seanwolter, code review?

PR #76, opened 2016-02-09
Parent/Child Contexts

chillpop:
Based on the guidelines from Marcus Zarra in this [blog post](http://martiancraft.com/blog/2015/03/core-data-stack/), I changed the way managed object contexts are setup to use a parent/child relationship.

There is now a private "root" context that saves to the persistent store coordinator.  The main context is a child of that root context.  Temporary contexts are children of the main context.  When saving, a context's parent contexts are also saved.

Although this doesn't affect the public interface much, it is a major change of how the stack is set up, so I've bumped the version to 3.0.

@vokal/ios-developers, @seanwolter code review?

I'm happy to add more tests to verify that the parent/child relationship works as expected if anyone has suggestions.


SwiftSampleProject/SwiftyVokoderTests/ManagedObjectContextTests.swift
```diff
@@ -0,0 +1,127 @@
...
+        let tempContext = manager.temporaryContext()
+        //temp context is a child of the main context
+        XCTAssertEqual(tempContext.parentContext, manager.managedObjectContext)
+        //main context has a private parent context
+        XCTAssertNotNil(manager.managedObjectContext.parentContext)
+    }
+    
+    func testDeletingObjectsOnTempContextGetsSavedToMainContext() {
+        //get a temp context, delete from temp, save to main, verify deleted on main
+        let manager = VOKCoreDataManager.sharedInstance()
```
bryanluby: Would it make sense to create properties out of `manager` and `tempContext` and do this configuration in `setUp`? - looks like it is pretty much the same for all tests

chillpop: :+1:

Pod/Classes/MapperMacros/VOKManagedObjectMapperMacros.h
```diff
@@ -39,36 +39,4 @@
 [VOKManagedObjectMap mapWithForeignKeyPath:inputKeyPath coreDataKey:VOKKeyForSelf(coreDataSelectorSymbol)]
 #endif
 
-
-////////////////////////// DEPRECATED MACROS //////////////////////////
```
bryanluby: were these recently deprecated?

chillpop: Yes, they were deprecated in #73 (v 2.4.1).

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -18,6 +18,8 @@ @interface VOKCoreDataManager ()
 @property (nonatomic, strong) NSManagedObjectModel *managedObjectModel;
 @property (nonatomic, strong) NSPersistentStoreCoordinator *persistentStoreCoordinator;
 @property (nonatomic, strong) NSManagedObjectContext *managedObjectContext;
+///A private parent context for writing to the persistent store
+@property (nonatomic, strong) NSManagedObjectContext *rootContext;
```
bryanluby: just my own feeling, but renaming this to something like `privateRootContext` would make this more self-documenting

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -484,78 +495,95 @@ - (NSManagedObjectContext *)safeContext:(NSManagedObjectContext *)context
...
+                  andWait:(BOOL)wait
+{
+    void (^saveBlock)() = ^{
+        NSError *error;
+        if ([context save:&error]) {
+            if (context.parentContext) {
+                [self saveContextToRoot:context.parentContext andWait:wait];
+            }
+        } else {
+            VOK_CDLog(@"Unresolved error %@, %@", error, [error localizedDescription]);
```
bryanluby: `localizedDescription` is a property

Pod/Classes/VOKCoreDataManager.m
brockboland: Seconded

Pod/Classes/MapperMacros/VOKManagedObjectMapperMacros.h
brockboland: byeeeeeee

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -90,7 +92,11 @@ - (NSManagedObjectContext *)managedObjectContext
     if (!_managedObjectContext) {
         NSAssert([NSOperationQueue currentQueue] == [NSOperationQueue mainQueue],
                  @"Must be on the main queue when initializing main context");
-        _managedObjectContext = [self managedObjectContextWithConcurrencyType:NSMainQueueConcurrencyType];
+        self.privateRootContext = [self managedObjectContextWithConcurrencyType:NSPrivateQueueConcurrencyType
+                                                           parentContext:nil];
```
brockboland: Colons got misaligned here with the longer property name

brockboland: Also: should this be moved into it's own lazy-loading accessor method? I could go either way on that one.

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -484,78 +495,95 @@ - (NSManagedObjectContext *)safeContext:(NSManagedObjectContext *)context
...
+ */
+- (void)saveContext:(NSManagedObjectContext *)context
+            andWait:(BOOL)wait
 {
-    [[NSNotificationCenter defaultCenter] addObserver:self
-                                             selector:@selector(tempContextSaved:)
-                                                 name:NSManagedObjectContextDidSaveNotification
-                                               object:tempContext];
+    __block BOOL hasChanges;
+    [context performBlockAndWait:^{
```
brockboland: Could you include a comment explaining why this is set inside a `performBlockAndWait`?

SwiftSampleProject/SwiftyVokoderTests/ManagedObjectContextTests.swift
```diff
@@ -0,0 +1,151 @@
...
+            object: self.manager.managedObjectContext) { _ in
+                
+                let newCountOfStations = self.manager.countForClass(Station.self)
+                XCTAssertEqual(newCountOfStations, 0)
+                XCTAssertNotEqual(countOfStations, newCountOfStations)
+                
+                return true
+        }
+        
+        if let rootContext = self.manager.managedObjectContext.parentContext {
```
brockboland: I'd make this a `guard let` and `XCTFail` in the `else` clause.

Pod/Classes/VOKCoreDataManager.m
chillpop: explain why `performBlockAndWait` instead of `performBlock` or explain why it's done using the block method at all?

brockboland: Why it's in a block at all. What I'm saying is: I don't know why you did this.

SwiftSampleProject/SwiftyVokoderTests/ManagedObjectContextTests.swift
```diff
@@ -0,0 +1,151 @@
...
+    func testUnsavedTempContextChangesDoNotGetPassedToMainContext() {
+        let countOfStations = self.manager.countForClass(Station.self)
+        XCTAssert(countOfStations > 0)
+
+        let childContext = self.manager.temporaryContext()
+        self.manager.deleteAllObjectsOfClass(Station.self, context: childContext)
+        
+        let childCountOfStations = self.manager.countForClass(Station.self, forContext: childContext)
+        XCTAssertNotEqual(countOfStations, childCountOfStations)
+        XCTAssertEqual(childCountOfStations, 0)
```
brockboland: Based on the name of this method, I would expect that it would check `self.manager.countForClass(Station.self)` again after deleting everything from the child context, to confirm that the count is **not** zero in the main context.

SwiftSampleProject/SwiftyVokoderTests/SwiftyVokoderTests.swift
```diff
@@ -12,8 +12,39 @@ import Vokoder
 
 let grandMilwaukeeIdentifier = 30096
 
+struct CTAData {
+
+    static func allStopDictionaries() -> [[String: AnyObject]] {
+        guard let
+            path = NSBundle.mainBundle().pathForResource("CTA_stations", ofType: "json"),
+            data = NSData(contentsOfFile: path)
+            else {
```
brockboland: Per coding standards, the `else` should be on the same line as the last condition

brockboland:
That's it from me

SwiftSampleProject/SwiftyVokoderTests/ManagedObjectContextTests.swift
chillpop: whoops, forgot that part

SwiftSampleProject/SwiftyVokoderTests/SwiftyVokoderTests.swift
```diff
@@ -12,8 +12,39 @@ import Vokoder
 
 let grandMilwaukeeIdentifier = 30096
 
+struct CTAData {
+
+    static func allStopDictionaries() -> [[String: AnyObject]] {
```
bryanluby: How bout adding a private `typealias` to this file to use for these:

```swift
private typealias JSONObject = [String: AnyObject]
```

chillpop:
updates pushed

SwiftSampleProject/SwiftyVokoderTests/SwiftyVokoderTests.swift
brockboland: I started doing this in another project and decided to remove it. Seeing `JSONObject` everywhere was just confusing because at a glance, it looked like it was supposed to be a class or struct somewhere.

bryanluby:
LGTM

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -484,78 +495,98 @@ - (NSManagedObjectContext *)safeContext:(NSManagedObjectContext *)context
...
-}
-
-- (void)tempContextSaved:(NSNotification *)notification
-{
-    // Solved issue with NSFetchedResultsController ignoring changes
-    // merged from different managed object contexts by touching
-    // willAccessValueForKey: on the updated objects.
-    //http://stackoverflow.com/questions/3923826/nsfetchedresultscontroller-with-predicate-ignores-changes-merged-from-different
+    [self obtainPermanentIDsForInsertionsInContext:context];
+    [self saveContextToRoot:context andWait:YES];
```
designatednerd: do you want to make the `wait` param here matched the passed-in param?

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -484,78 +495,98 @@ - (NSManagedObjectContext *)safeContext:(NSManagedObjectContext *)context
...
+        } else {
+            VOK_CDLog(@"Unresolved error %@, %@", error, error.localizedDescription);
+        }
+    };
     
-    for (NSManagedObject *object in [[notification userInfo] objectForKey:NSUpdatedObjectsKey]) {
-        [[self.managedObjectContext objectWithID:[object objectID]] willAccessValueForKey:nil];
+    //the root context can be saved asynchronously
+    if (context == self.privateRootContext) {
+        wait = NO;
```
designatednerd: This seems a bit weird - this is basically saying no matter what, don't wait if it's the root context. I'm not sure I'd expect that behavior as a caller.

Pod/Classes/VOKCoreDataManager.m
chillpop: yes, it should

SwiftSampleProject/SwiftyVokoderTests/ManagedObjectContextTests.swift
```diff
@@ -0,0 +1,156 @@
...
+    
+    func testUnsavedMainContextChangesGetPassedToTempContexts() {
+        let countOfStations = self.manager.countForClass(Station.self)
+        XCTAssert(countOfStations > 0)
+
+        let childContextBeforeChanges = self.manager.temporaryContext()
+        self.manager.deleteAllObjectsOfClass(Station.self, context: nil)
+        let childContextAfterChanges = self.manager.temporaryContext()
+
+        let childCountOfStations = self.manager.countForClass(Station.self, forContext: childContextBeforeChanges)
```
designatednerd: can you add a bit of documentation about exactly what this is testing - it took me a second to figure out that the two `temporaryContext` calls wouldn't be pointed at the same instance

SwiftSampleProject/SwiftyVokoderTests/ManagedObjectContextTests.swift
```diff
@@ -0,0 +1,156 @@
...
+        XCTAssertNotEqual(countOfStations, childCountOfStations)
+        XCTAssertEqual(childCountOfStations, 0)
+        
+        let newChildCountOfStations = self.manager.countForClass(Station.self, forContext: childContextAfterChanges)
```
designatednerd: I feel like this var name violates the `NARC` rule

SwiftSampleProject/SwiftyVokoderTests/ManagedObjectContextTests.swift
```diff
@@ -0,0 +1,156 @@
...
+        XCTAssert(countOfStations > 0)
+
+        let childContext = self.manager.temporaryContext()
+        self.manager.deleteAllObjectsOfClass(Station.self, context: childContext)
+        
+        let childCountOfStations = self.manager.countForClass(Station.self, forContext: childContext)
+        XCTAssertNotEqual(countOfStations, childCountOfStations)
+        XCTAssertEqual(childCountOfStations, 0)
+        
+        let newCountOfStations = self.manager.countForClass(Station.self)
```
designatednerd: (same here on `NARC` - i usually use `updated` instead of `new` to distinguish)

SwiftSampleProject/SwiftyVokoderTests/ManagedObjectContextTests.swift
chillpop: Isn't that only applied to methods?

designatednerd: [It definitely applies to methods](https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/mmRules.html). There's no mention of variable naming in that doc, though I have a sense memory of being smacked on this in code review a while back for variable naming as well.

tl;dr You appear to be correct, but I feel like it still looks weird if you're using `new` without actually creating something.

Pod/Classes/VOKCoreDataManager.m
chillpop: At this point, the main context is already saved.  If you're using Vokoder and making changes through the main context and any temp contexts, those changes will be reflected in the contexts you have.  (If, on the other hand, you're **abusing** Vokoder and using the private root context directly, you won't see changes to the root context immediately.)

Perhaps this should be made explicit in the README and the comments.

designatednerd: Definitely think README update would be helpful.

chillpop:
updates pushed

designatednerd:
LGTM pending big T.

SwiftSampleProject/SwiftyVokoderTests/ManagedObjectContextTests.swift
```diff
@@ -0,0 +1,160 @@
...
+        //get a temp context, delete from temp, save to main, verify deleted on main
+        
+        let tempContext = self.manager.temporaryContext()
+        
+        let countOfStations = self.manager.countForClass(Station.self)
+        XCTAssert(countOfStations > 0)
+        
+        tempContext.performBlockAndWait {
+            self.manager.deleteAllObjectsOfClass(Station.self, context: tempContext)
+            self.manager.saveAndMergeWithMainContextAndWait(tempContext)
```
brockboland: Should the `saveAndMergeWithMainContextAndWait` method normally be invoked from a block on the context itself? If so, should we document it as such? I skimmed the doc changes and didn't see this clarified.

brockboland: And/or: maybe we should clarify when it's appropriate to use this approach as opposed to the `writeToTemporaryContext` class method

chillpop: As the README says, `writeToTemporaryContext` is more of a convenience method: it will handle creating and saving the context for you.  The `saveAndMergeWithMainContext...` methods are for when you've already created a temp context yourself and want to save and merge changes back to main context (and persistent store).  

The `saveAndMergeWithMainContext...` methods are safe to call from any queue (and described as such in the header docs) because all of the context stuff is wrapped in `performBlock...` calls (like the `hasChanges` call you asked about earlier).

brockboland: Got it. But that means your call to `saveAndMergeWithMainContextAndWait` here doesn't need to happen within the block, since that method itself winds up running in a `performBlock` call anyway, right?

chillpop: You're right.  I was moving those lines around to see if it made a difference and I forgot to move them back.

chillpop: originally, I had them outside the blocks: https://github.com/chillpop/Vokoder/blob/1765d3b9b08c1719a9c08d1efc3ba8a5d936f9fd/SwiftSampleProject/SwiftyVokoderTests/ManagedObjectContextTests.swift#L43-L47

brockboland: That's why I asked about this one: I saw the diff where you moved this inside the block, and wasn't sure if that was how these should be handled across the board.

brockboland:
LGTM! Any final input @designatednerd @bryanluby?

bryanluby:
LGTM

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -484,78 +495,100 @@ - (NSManagedObjectContext *)safeContext:(NSManagedObjectContext *)context
...
+    //The documentation indicates that this method is not thread safe.
+    //We can't guarantee that the current method is called on the appropriate thread for this context,
+    //so use performBlockAndWait to query the context for changes.
+    __block BOOL hasChanges;
+    [context performBlockAndWait:^{
+        hasChanges = context.hasChanges;
+    }];
     
-    [self saveContext:tempContext];
+    if (!hasChanges) {
```
seanwolter: I always wondered if checking for changes before saving was a premature optimization. From the docs:
"Always verify that the context has uncommitted changes (using the hasChanges property) before invoking the save: method. Otherwise, Core Data may perform unnecessary work."
Now I know.

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -484,78 +495,100 @@ - (NSManagedObjectContext *)safeContext:(NSManagedObjectContext *)context
...
-                                                  object:tempContext];
-}
-
-- (void)tempContextSaved:(NSNotification *)notification
-{
-    // Solved issue with NSFetchedResultsController ignoring changes
-    // merged from different managed object contexts by touching
-    // willAccessValueForKey: on the updated objects.
-    //http://stackoverflow.com/questions/3923826/nsfetchedresultscontroller-with-predicate-ignores-changes-merged-from-different
+    [self obtainPermanentIDsForInsertionsInContext:context];
```
seanwolter: Why is it necessary to `obtainPermanentIDsForInsertionsInContext:` before saving? (Sorry if I missed an earlier discussion)

brockboland: `obtainPermanentIDsForObjects` was being called directly before, further down in the code. He just refactored some things around. I understood why this is needed yesterday, but not well enough to explain it today.

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -484,78 +495,100 @@ - (NSManagedObjectContext *)safeContext:(NSManagedObjectContext *)context
...
+        }
+    };
     
-    for (NSManagedObject *object in [[notification userInfo] objectForKey:NSUpdatedObjectsKey]) {
-        [[self.managedObjectContext objectWithID:[object objectID]] willAccessValueForKey:nil];
+    //At this point, the main context is already saved.
+    //The root context can be saved asynchronously
+    //since all access to Core Data should be going through the main context or a child temp context.
+    if (context == self.privateRootContext) {
+        wait = NO;
```
seanwolter: Is this supposed to change `wait` to NO for the `saveBlock`? Does that work? I thought blocks captured variables at their time of creation unless `__block` is specified.

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -484,78 +495,100 @@ - (NSManagedObjectContext *)safeContext:(NSManagedObjectContext *)context
...
     }
     
-    if ([NSOperationQueue mainQueue] == [NSOperationQueue currentQueue]) {
-        [self.managedObjectContext mergeChangesFromContextDidSaveNotification:notification];
+    if (wait) {
```
seanwolter: oh I see, please disregard my previous comment

seanwolter:
LGTM

Pod/Classes/VOKCoreDataManager.m
designatednerd: giphy

designatednerd:
SLGTM

Pod/Classes/MapperMacros/VOKManagedObjectMapperMacros.h
```diff
@@ -39,36 +39,4 @@
...
- *  Creates a map with the default date mapper.
- *
- *  @param inputKeyPath           The foreign key to match with the local key.
- *  @param coreDataSelectorSymbol The local selector symbol.
- *
- *  @return A VOKManagedObjectMap
- */
-#ifndef VOK_MAP_FOREIGN_TO_LOCAL
-#   define VOK_MAP_FOREIGN_TO_LOCAL(inputKeyPath, coreDataSelectorSymbol) \
-_Pragma ("GCC warning \"'VOK_MAP_FOREIGN_TO_LOCAL' macro is deprecated and will be removed in version 3.0. Use VOKMapForeignToLocalClassProperty or VOKMapForeignToLocalForSelf instead.\"") \
```
vokal-isaac: Bye, clever macro-deprecating trick from http://stackoverflow.com/a/29297970/291280

Pod/Classes/VOKCoreDataManager.h
```diff
@@ -243,25 +245,27 @@ typedef void(^VOKObjectIDsReturnBlock)(VOKArrayOfManagedObjectIDs *managedObject
...
 /**
- Saves the main context synchronously on the main queue. If already on the main queue it performs the same as saveMainContext.
+ Saves the main context synchronously. This method will not return until the main context is saved.
  */
 - (void)saveMainContextAndWait;
 
 /**
- Provides a managed object context for scratch work or background processing using the same persistent store coordinator as the main context. As with all managed object contexts, it is not thread-safe.
+ Provides a managed object context for scratch work or background processing as a child of the main context. As with all managed object contexts, it is not thread-safe.
  Create the context and do work on the same queue. You are responsible for retaining temporary contexts yourself.
```
vokal-isaac: If the idea is that this is a private-queue context, then I think the thread/queue it's created on doesn't so much matter as only using it via `performBlock:` and `performBlockAndWait:`.

Pod/Classes/VOKCoreDataManager.h
```diff
@@ -243,25 +245,27 @@ typedef void(^VOKObjectIDsReturnBlock)(VOKArrayOfManagedObjectIDs *managedObject
...
 
 /**
- Saves the main context synchronously on the main queue. If already on the main queue it performs the same as saveMainContext.
+ Saves the main context synchronously. This method will not return until the main context is saved.
  */
 - (void)saveMainContextAndWait;
 
 /**
- Provides a managed object context for scratch work or background processing using the same persistent store coordinator as the main context. As with all managed object contexts, it is not thread-safe.
+ Provides a managed object context for scratch work or background processing as a child of the main context. As with all managed object contexts, it is not thread-safe.
```
vokal-isaac: If this is a child of the main context, then I think saving this context is going to cause the main context to save, tying those threads/queues together.  Is that the intended behavior or should this be a child of the private root context?

Pod/Classes/VOKCoreDataManager.h
chillpop: yes, good point

Pod/Classes/VOKCoreDataManager.h
chillpop: If you save a child context, the changes only get saved up to the immediate parent context.  The parent context changes do not automatically get saved too.  (That is the reason for the recursive `saveContextToRoot:` method.)

Pod/Classes/VOKCoreDataManager.h
```diff
@@ -285,13 +289,20 @@ typedef void(^VOKObjectIDsReturnBlock)(VOKArrayOfManagedObjectIDs *managedObject
...
- Saves any temporary managed object context and merges those changes with the main managed object context in a thread-safe way.
+ Saves any temporary managed object context and merges those changes with the main managed object context asynchronously.
  This method is safe to call from any queue.
- @param context     The termporary context to save. Do not use this method to save the main context.
+ @param context     The temporary context to save.
  */
 - (void)saveAndMergeWithMainContext:(NSManagedObjectContext *)context;
 
 /**
+ Saves any temporary managed object context and merges those changes with the main managed object context synchronously.
```
vokal-isaac: Similar concern here—it sounds like this method ties an action in one context to an action in another, tying their queues together and potentially risking a deadlock.

Pod/Classes/VOKCoreDataManager.h
vokal-isaac: Ahh, hmm... so saving the temporary context will cause those changes to be made in the main context... which has to happen on the main context's queue, right?  I _think_ that linking of execution across queues is how I wound up with deadlocking issues when I last used parent-child contexts.

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -90,7 +92,11 @@ - (NSManagedObjectContext *)managedObjectContext
     if (!_managedObjectContext) {
         NSAssert([NSOperationQueue currentQueue] == [NSOperationQueue mainQueue],
                  @"Must be on the main queue when initializing main context");
```
vokal-isaac: Is this still true?

chillpop: hmm, probably not

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -204,15 +210,20 @@ - (void)addPersistentStoreToCoordinator:(NSPersistentStoreCoordinator *)persiste
...
-    NSPersistentStoreCoordinator *coordinator = self.persistentStoreCoordinator;
-    
-    NSAssert(coordinator, @"PersistentStoreCoordinator does not exist. This is a big problem.");
-    if (!coordinator) {
-        return nil;
-    }
     NSManagedObjectContext *context = [[NSManagedObjectContext alloc] initWithConcurrencyType:concurrencyType];
-    context.persistentStoreCoordinator = coordinator;
+    if (parentContext) {
+        context.parentContext = parentContext;
```
vokal-isaac: [The docs](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/CoreDataFramework/Classes/NSManagedObjectContext_Class/#//apple_ref/doc/uid/TP30001182-256616) say

> As the first scenario implies, a parent context can service requests from children on different threads. You cannot, therefore, use parent contexts created with the thread confinement type

Is it worth an assertion that `parentContext.concurrencyType != NSConfinementConcurrencyType`?

Pod/Classes/VOKCoreDataManager.h
chillpop: All of the saves (when done through Vokoder methods) happen inside a context's `performBlock...:` methods which are intended to work around threading issues.

Pod/Classes/MapperMacros/VOKManagedObjectMapperMacros.h
seanwolter: byeeeeeee

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -484,78 +495,100 @@ - (NSManagedObjectContext *)safeContext:(NSManagedObjectContext *)context
...
+- (void)saveContextToRoot:(NSManagedObjectContext *)context
+                  andWait:(BOOL)wait
+{
+    void (^saveBlock)() = ^{
+        NSError *error;
+        if ([context save:&error]) {
+            if (context.parentContext) {
+                [self saveContextToRoot:context.parentContext andWait:wait];
+            }
+        } else {
```
vokal-isaac: If I understand this correctly (that's a big _if_), if one were to save a temporary context with `wait == YES`, the temporary context will perform-and-wait with this block, which will save the parent context (the main context) with perform-and-wait, blocking the main queue, which will save the private root context with perform-and-wait, writing to disk, so the main queue will be blocked while the changes are saved to disk.

vokal-isaac: Ugh, I missed the forced-toggling of `wait` to `NO`... I think I'm probably wrong...

Pod/Classes/VOKCoreDataManager.m
chillpop: There's already an exception thrown if that happens so I don't think it's necessary.

vokal-isaac: Ahh, if doing that throws an exception, then that's fine.

Pod/Classes/VOKCoreDataManager.m
chillpop: Yes, the root context is saved asynchronously on it's private queue, so the main queue is only blocked while the main context is being saved (and `wait == YES`).

vokal-isaac: I think calling `saveAndMergeWithMainContextAndWait:` on the main queue with a temporary context will cause a deadlock?

chillpop: based on the new tests, I don't think so:
https://github.com/vokal/Vokoder/pull/76/files#diff-9c539a463d05728d06f86ae9970392f5R64

vokal-isaac: Given our discussion, experimentation around that test to which you linked, and http://stackoverflow.com/a/11833139/291280, it seems that it's much harder to deadlock `performBlockAndWait:` than I'd thought.

Pod/Classes/VOKCoreDataManager.h
```diff
@@ -243,31 +245,33 @@ typedef void(^VOKObjectIDsReturnBlock)(VOKArrayOfManagedObjectIDs *managedObject
...
  @endcode
- @return        A managed object context with the same persistent store coordinator as the main context, but otherwise no relationship.
+ @return        A managed object context with the main context as its parent.
  */
 - (NSManagedObjectContext *)temporaryContext;
 
 /**
  This provides a way for an application with heavy amounts of Core Data threading and writing to maintain object graph integrety by assuring that only one context is being written to at once.
- @param writeBlock      Do not save or merge this context, it will be done for you.  Do not use GCD or thread jumping inside this block. 
+ @param writeBlock      Do not save or merge this context, it will be done for you.
```
vokal-isaac: I'd think the no-GCD-no-thread-jumping still applies?

Pod/Classes/VOKCoreDataManager.h
vokal-isaac: Nevermind.

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -484,78 +491,106 @@ - (NSManagedObjectContext *)safeContext:(NSManagedObjectContext *)context
...
+                                           parentContext:self.managedObjectContext];
 }
 
 - (void)saveAndMergeWithMainContext:(NSManagedObjectContext *)context
 {
-    NSAssert(context != self.managedObjectContext, @"This is NOT for saving the main context.");
-    [self saveTempContext:context];
+    NSAssert(context.parentContext != nil,
+             @"%@ is for saving temp contexts that are descendents of the main context!",
+             NSStringFromSelector(_cmd));
```
vokal-isaac: Worth a check that the main context is somewhere in `context`'s lineage?

README.md
```diff
@@ -148,7 +148,7 @@ The mapper constructed in the example in the section above could be included in
 
 ###Importing Safely
 
-Vokoder offers many ways to get data into Core Data. The simplest and most approachable interface is offered through the VOKManagedObjectAdditions category. Given an array of dictionaries Vokoder will create or edit managed objects on a background queue and then safely return managed objects to the main queue through the completion block.
+Vokoder offers many ways to get data into Core Data. The simplest and most approachable interface is offered through the VOKManagedObjectAdditions category. Given an array of dictionaries, Vokoder will create or update managed objects on a temporary context and then safely return managed objects from the main context through a completion block on the main queue.
```
vokal-isaac: Looking at the implementation of `vok_addWithArrayInBackground:completion:`, I'm wondering if it should use `performBlockAndWait:`?

SampleProject/VOKCoreDataManagerTests/VOKManagedObjectAdditions.m
```diff
@@ -156,17 +154,17 @@ - (void)testRecordInsertionBackgroundThreadConvenienceReturningManagedObjectsArr
                                  @"numberOfHats" : @1,
                                  },
                              ];
-
+    
     [VOKThing vok_addWithArrayInBackground:importArray
-                               completion:^(NSArray *arrayOfManagedObjects) {
```
vokal-isaac: Probably a future thing and not for this PR, but could `vok_addWithArrayInBackground:completion:` be set up so that the completion parameter was `NSArray<instancetype> *`? (I'm pretty sure `instancetype` can't be used in that way, but maybe achieve that effect through some other means?)

SampleProject/VOKCoreDataManagerTests/VOKManagedObjectAdditions.m
```diff
@@ -156,17 +154,17 @@ - (void)testRecordInsertionBackgroundThreadConvenienceReturningManagedObjectsArr
...
-                                       XCTAssertEqualObjects([arrayOfManagedObjects[i] numberOfHats], [importArray[i] valueForKey:@"numberOfHats"]);
-                                   }
-                                   [completionHandlerExpectation fulfill];
-    }];
-
+                                completion:^(NSArray *arrayOfManagedObjects) {
+                                    XCTAssertEqual(arrayOfManagedObjects.count, importArray.count);
+                                    for (NSInteger i = 0; i < importArray.count; i++) {
+                                        XCTAssertEqualObjects([arrayOfManagedObjects[i] name], [importArray[i] valueForKey:@"name"]);
+                                        XCTAssertEqualObjects([arrayOfManagedObjects[i] numberOfHats], [importArray[i] valueForKey:@"numberOfHats"]);
```
vokal-isaac: Worth using macros instead of string keys?

SwiftSampleProject/SwiftyVokoderTests/ManagedObjectContextTests.swift
```diff
@@ -0,0 +1,160 @@
...
+        XCTAssertNotNil(self.manager.managedObjectContext.parentContext)
+    }
+    
+    func testDeletingObjectsOnTempContextGetsSavedToMainContext() {
+        //get a temp context, delete from temp, save to main, verify deleted on main
+        
+        let tempContext = self.manager.temporaryContext()
+        
+        let countOfStations = self.manager.countForClass(Station.self)
+        XCTAssert(countOfStations > 0)
```
vokal-isaac: There's a `XCTAssertGreaterThan`.

Not saying you should use it; just mentioning that it exists.

vokal-isaac:
That's it for me.

README.md
chillpop: Under the hood, `vok_addWithArrayInBackground:completion` uses `importArrayInBakcground:forClass:completion:`. If there is a completion block passed in, it saves synchronously: https://github.com/vokal/Vokoder/pull/76/files#diff-95e5223a9d0ae5969d41279c482be816R626

SampleProject/VOKCoreDataManagerTests/VOKManagedObjectAdditions.m
chillpop: I don't think you can use `instancetype` as an Obj-C generic type.  As it is now, that method returns `NSArray<__kindof NSManagedObject *> *` in Obj-C.  

With the Swift additions, you **can** get a properly typed array of your managed object subclass either through the generic [`importArray`](https://github.com/vokal/Vokoder/blob/master/Pod/Classes/Swift/VOKCoreDataManager%2BSwift.swift#L37-L46) function, or by adopting the `VokoderTypedManagedObject` protocol and using the [`vok_import`](https://github.com/vokal/Vokoder/blob/master/Pod/Classes/Swift/VokoderTypedManagedObject.swift#L14-L33) function.

SampleProject/VOKCoreDataManagerTests/VOKManagedObjectAdditions.m
chillpop: sure

SwiftSampleProject/SwiftyVokoderTests/ManagedObjectContextTests.swift
chillpop: good to remember

SampleProject/VOKCoreDataManagerTests/VOKManagedObjectAdditions.m
chillpop: Actually, it should be using subscript syntax: `importArray` is an array of dictionaries.

chillpop:
updates pushed

README.md
vokal-isaac: I should have been more specific—I meant inside the completion block that's passed to `importArrayInBackground:forClass:completion:`.

brockboland:
SLGTM. Does that address all of your concerns, @vokal-isaac?

vokal-isaac:
I may have missed it, but I think these two haven't been addressed:

- https://github.com/vokal/Vokoder/pull/76#discussion_r52648033
- https://github.com/vokal/Vokoder/pull/76#discussion_r52670010

vokal-isaac:
LGTM (discussed in person).

chillpop:
merged, tagged, and `pod trunk push`'d

PR #77, opened 2016-02-29
Documentation updates and minor code modernizations

brockboland:
This one looks worse than it is…I think. I updated a bunch of comments and improved the README, and tweaked a few things in the process (like asserting the main thread and switching to `instancetype` as the return type on init methods).

@vokal/ios-developers this is a lower-priority PR to review, please


naomihimley:
I will look but someone else with more Vokoder understanding should look as well

Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.h
```diff
@@ -18,21 +18,21 @@ typedef void(^VOKManagedObjectsReturnBlock)(VOKArrayOfManagedObjects *managedObj
 @interface NSManagedObject (VOKManagedObjectAdditions)
 
 /**
- Checks for NSNull before seting a value.
- @param value   The input object.
+ Checks for NSNull before seting a value on this object.
```
naomihimley: might as well fix the seting -> setting spelling here as well

Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.h
```diff
@@ -148,10 +148,10 @@ typedef void(^VOKManagedObjectsReturnBlock)(VOKArrayOfManagedObjects *managedObj
                                forManagedObjectContext:(nullable NSManagedObjectContext *)contextOrNil;
 
 /*
- Returns one entite matching the predicate. Asserts the count is exactly 1. If more objects are returned this method will let you know.
+ Returns one entity matching the predicate. Asserts the count is exactly 1. If more objects are returned, the last one returned by Core Data will be returned.
```
naomihimley: It is correct that the last one is returned but I don't know that it makes sense to remove the part that tells you that it hits the assert if so. I'm fine with it though, since it says "Asserts the count is exactly 1" before

Pod/Classes/Optional Data Sources/VOKFetchedResultsDataSource.h
```diff
@@ -13,82 +13,225 @@ NS_ASSUME_NONNULL_BEGIN
...
 
+/**
+ Accessor for the fetchedObjects array on the fetched results controller
+
+ @return Array of fetched NSManagedObjects
+ */
 - (nullable VOKArrayOfManagedObjects *)fetchedObjects;
 
+/**
+ Reload the fetcehd data in the fetched results controller, which will cause a table view reload.
```
naomihimley: fetched misspelled

Pod/Classes/Optional Data Sources/VOKFetchedResultsDataSource.h
```diff
@@ -13,82 +13,225 @@ NS_ASSUME_NONNULL_BEGIN
...
-              cacheName:(nullable NSString *)cacheName
-              tableView:(nullable UITableView *)tableView
-     sectionNameKeyPath:(nullable NSString *)sectionNameKeyPath
-        sortDescriptors:(nullable VOKArrayOfSortDescriptors *)sortDescriptors
-     managedObjectClass:(Class)managedObjectClass
-              batchSize:(NSInteger)batchSize
-             fetchLimit:(NSInteger)fetchLimit
-               delegate:(nullable id <VOKFetchedResultsDataSourceDelegate>)delegate;
+/**
+ Reload the fetcehd data in the fetched results controller, which will cause a table view reload.
```
naomihimley: misspelled here as well. maybe you should just stop trying to make fetch happen...

Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.h
brockboland: That's part of the reason I removed it: "this method will let you know" isn't really descriptive, and it already explains the assertion. I just wanted to clarify what would happen in a release build.

Pod/Classes/Optional Data Sources/VOKFetchedResultsDataSource.h
brockboland: I'm trying to make fetceh happen

Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.h
```diff
@@ -45,14 +45,14 @@ typedef void(^VOKManagedObjectsReturnBlock)(VOKArrayOfManagedObjects *managedObj
...
 + (instancetype)vok_newInstance;
 
 /**
- A convenience methods to create a new instance of a NSManagedObject subclass.
- @param contextOrNil    The managed object context to insert the new object.  If nil, the main context will be used.
+ A convenience method to create a new instance of a NSManagedObject subclass.
+ @param contextOrNil    The managed object context in which to insert the new object.  If nil, the main context will be used.
  @return                A new managed object subclass in the main context.
  */
 + (instancetype)vok_newInstanceWithContext:(nullable NSManagedObjectContext *)contextOrNil;
```
chillpop: At some point, we might want to standardize the preposition we use to indicate the managed object context.  "with context", "for context", and "in context" being the three options that I think I've seen.

seanwolter:
This might be a good time to update the copyright to 2016

Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.h
```diff
@@ -61,31 +61,31 @@ typedef void(^VOKManagedObjectsReturnBlock)(VOKArrayOfManagedObjects *managedObj
...
- @param contextOfNil    The managed object context to update and/or insert the object. If nil, the main context will be used.
+ @param contextOfNil    The managed object context in which to update and/or insert the object. If nil, the main context will be used.
  @return                An instance of this subclass of NSManagedObject.
  **/
 + (nullable instancetype)vok_addWithDictionary:(VOKStringToObjectDictionary *)inputDict
                        forManagedObjectContext:(nullable NSManagedObjectContext *)contextOrNil;
 
 /*
- Create or update many NSManagedObjects, respecting the mapper's overwriteObjectsWithServerChanges and ignoreNullValueOverwrites properties.
+ Create or update many NSManagedObjects in a background managed object context, respecting the mapper's overwriteObjectsWithServerChanges and ignoreNullValueOverwrites properties.
```
chillpop: How about "on a background operation queue"?  I'm worried that "background managed object context" is unclear/inaccurate.

Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.h
```diff
@@ -61,31 +61,31 @@ typedef void(^VOKManagedObjectsReturnBlock)(VOKArrayOfManagedObjects *managedObj
...
  This should only be used to set all properties of an entity.
  By default any mapped attributes not included in the input dictionaries will be set to nil.
- This will overwrite ALL of an NSManagedObject's properties unless ignoreNullValueOverwrites is YES;
+ This will overwrite ALL of an NSManagedObject's properties unless ignoreNullValueOverwrites is YES on the mapper for this class,
```
chillpop: no need for the comma at the end

Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.h
chillpop: how about adding that the "If more objects are returned..." only applies if assertions are turned off?

Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.h
```diff
@@ -148,10 +148,10 @@ typedef void(^VOKManagedObjectsReturnBlock)(VOKArrayOfManagedObjects *managedObj
                                forManagedObjectContext:(nullable NSManagedObjectContext *)contextOrNil;
 
 /*
- Returns one entite matching the predicate. Asserts the count is exactly 1. If more objects are returned this method will let you know.
+ Returns one entity matching the predicate. Asserts the count is exactly 1. If more objects are returned, the last one returned by Core Data will be returned.
  @param predicate       Predicate to use to fetch.
  @param contextOrNil    The managed object context to fetch in.  If nil, the main context will be used.
```
chillpop: "in which to fetch"?

brockboland:
@seanwolter I did some searching and the consensus seems to be that copyright date should indicate when it was written, as the copyright extends from that time. I guess we should not have updated those dates in the past like we did.

Pod/Classes/Optional Data Sources/VOKCollectionDataSource.h
```diff
@@ -9,51 +10,124 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/**
+ A generic data source and delegate for collection views, backed by a NSFetchedResultsController.
```
chillpop: **an** NSFetchedResultsController?

Pod/Classes/Optional Data Sources/VOKCollectionDataSource.h
```diff
@@ -9,51 +10,124 @@
...
+/**
+ Setup this data source.
+
+ @param predicate          Predicate used to filter the objects displayed
+ @param cacheName          Name for the fetched results controller cache
+ @param collectionView     Collection view in which results are shown. The data source and delegate are set to the newly-created data source.
+ @param sectionNameKeyPath Keypath to use to group results in sections
+ @param sortDescriptors    Sort descriptors to sort the objects
+ @param managedObjectClass NSManagedObject subclass to fetch
+ @param batchSize          Batch size to use for the fetch request fetchBatchSize. Defaults to 20.
```
chillpop: I think it makes more sense to put the default value comment on methods that _don't_ include the `batchSize` parameter.

Pod/Classes/Optional Data Sources/VOKCollectionDataSource.h
```diff
@@ -9,51 +10,124 @@
...
+ @param delegate           Delegate to notify of cell selection/deselection and whether the fetch has results
```
chillpop: technically, the delegate gets notified of _object_ selection, not cell selection.

Pod/Classes/VOKCoreDataManager.h
```diff
@@ -57,7 +58,7 @@ typedef void(^VOKObjectIDsReturnBlock)(VOKArrayOfManagedObjectIDs *managedObject
 @property (nonatomic, strong, readonly) NSManagedObjectModel *managedObjectModel;
 
 /**
- Set the name of the managed object model and the name of the SQL lite store on disk. Call this first when you setup the core data stack.
+ Set the name of the managed object model and the name of the SQL lite store on disk. Call this first when you setup the Core Data stack.
```
chillpop: "SQL lite" -> SQLite

Pod/Classes/VOKCoreDataManager.h
```diff
@@ -67,18 +68,19 @@ typedef void(^VOKObjectIDsReturnBlock)(VOKArrayOfManagedObjectIDs *managedObject
            database:(nullable NSString *)database;
 
 /**
- Set the name of the managed object model and the name of the SQL lite store on disk, in the provided bundle. Call this first when you setup the core data stack.
+ Set the name of the managed object model and the name of the SQL lite store on disk, in the provided bundle. Call this first when you setup the Core Data stack.
```
chillpop: SQLite here too

Pod/Classes/VOKCoreDataManager.h
```diff
@@ -67,18 +68,19 @@ typedef void(^VOKObjectIDsReturnBlock)(VOKArrayOfManagedObjectIDs *managedObject
...
+ @param bundle      The bundle where the Managed Object Model can be found. If nil, the main bundle will be used.
  */
 - (void)setResource:(nullable NSString *)resource
            database:(nullable NSString *)database
              bundle:(nullable NSBundle *)bundle;
 
 /**
- In case of a migration failure, these options allow possible recovery and notification
+ In case of a migration failure, these options allow possible recovery and notification.
+ Defaults to VOKMigrationFailureOptionNone/
```
chillpop: there's a trailing /

Pod/Classes/Optional Data Sources/VOKCollectionDataSource.h
brockboland: Good point. The default is mentioned on the property declaration so I'll leave it at that.

README.md
```diff
@@ -173,55 +178,62 @@ NSManagedObjectContext *backgroundContext = [[VOKCoreDataManager sharedInstance]
...
-[person setFirstName:@"Rohan"];
-[person setLastName:@"Panchal"];
+person.firstName = @"Rohan";
+person.lastName = @"Panchal";
 [[VOKCoreDataManager sharedInstance] saveMainContextAndWait];
 ```
 
-###Querying Records	
+###Querying Records
+Like the example above, this one makes use of the `VOKKeyForInstanceOf` macro from [VOKUtilities/VOKKeyPathHelper](https://github.com/vokal/VOKUtilities#vokkeypathhelper) for the keys in these calls to ensure that typos like `@"lsatName"` don't occur.
```
chillpop: `lsatName` -> `lastName`
We're not applying to law school here

chillpop: oh, I fail at reading comprehension

README.md
```diff
@@ -173,55 +178,62 @@ NSManagedObjectContext *backgroundContext = [[VOKCoreDataManager sharedInstance]
...
 
 ####Query with basic predicate
 ```objective-c
-NSArray *results = [VOKPerson vok_fetchAllForPredicate:nil forManagedObjectContext:nil]; //Basic Fetch
+    NSPredicate *smithsPredicate = [NSPredicate predicateWithFormat:@"%K == %@"
+                                                      argumentArray:@[VOKKeyForInstanceOf([VOKPerson class], lastName), @"Smith"]];
```
chillpop: 1. Should be `VOKKeyForInstanceOf(VOKPerson, lastName)`.  The macro calls `class` for you, (also below).
1. Any reason to prefer using `argumentsArray` instead of var args?

SampleProject/VOKCoreDataManagerTests/VOKThreadingTests.m
```diff
@@ -0,0 +1,37 @@
...
+
+    // Setup on the main thread should work
+    [[VOKCoreDataManager sharedInstance] resetCoreData];
+    XCTAssertNoThrow([[VOKCoreDataManager sharedInstance] setResource:@"VOKCoreDataModel" database:nil]);
+
+    NSOperationQueue *backgroundQueue = [[NSOperationQueue alloc] init];
+    backgroundQueue.qualityOfService = NSOperationQualityOfServiceBackground;
+    [backgroundQueue addOperationWithBlock:^{
+        [[VOKCoreDataManager sharedInstance] resetCoreData];
+        // Setup on a background thread should fail
```
chillpop: Without the new assert you put in, does this test fail?  i.e.: Does the context actually need to be created on the main queue due to internal Core Data stuff?

README.md
brockboland: 1. Whoops
2. Been away from Objective-C for a bit

Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.h
brockboland: https://github.com/vokal/Vokoder/issues/78

SampleProject/VOKCoreDataManagerTests/VOKThreadingTests.m
brockboland: The test fails because of the `XCTAssertThrows`, but to your real question: if I remove the assertions in VOKCoreDataManager (including the existing one), and then use this background thread to setup the MOC and create an object, CoreData seems to be OK. Same is true if I create the MOC on the main thread and then create an object in the background thread. I think problems will only surface later on in more complicated, real-world situations that I'm not sure how to test well, which is why I added this better-than-nothing-but-not-great test.

chillpop: If you can create the MOC without problem on a background thread, I'd argue that the assertion you added isn't needed.  If you use the main context directly on the main thread and use the `performBlock` methods when on a different thread things should work. If you're worried about problems surfacing later, I don't know if the main queue assertion will help. Do you have an example of problems or just an abundance of caution?

brockboland: An abundance of caution and this comment on `setResource:database`:

> The main context will be initialized immediately. This method should only be called from the main queue

That was added by @seanwolter in 9954a8909d9a7f3cd2f4c1f4bf44143b02ce789f but I can't find the PR with the feedback that prompted it.

chillpop: Previously, it was important that the main MOC be created on the main queue.  Using the new(-ish) `NSMainQueueConcurrencyType` and `NSPrivateQueueConcurrencyType` options, it shouldn't matter on what queue the MOC is created, just how you use it.  I'd recommend removing the comments that suggest otherwise.

brockboland: Those have been around since iOS 5, so why was that warning added to the comments last May? Lingering superstition? Or did #72 and #76 modernize things enough in the interim such that said warning can be removed? If so, I'm assuming that the [other main queue assertion](https://github.com/vokal/Vokoder/blob/master/Pod/Classes/VOKCoreDataManager.m#L484) can be dropped as well.

chillpop: I think it's just lingering suspicion and/or making the comments line up with the internal assertions in the code.
I think that other assertion should remain though.  The methods that use `safeContext:` aren't wrapping their calls to the context inside of `performBlock` so they should be called on the main queue.

brockboland:
@chillpop updated

chillpop:
LGTM

brockboland:
Anybody else want to have a look? @vokal/ios-developers @seanwolter @designatednerd (just in case)

seanwolter:
It *just* works!
lgtm

designatednerd:
LGTM :v:

PR #81, opened 2016-04-27
Specify locale to ensure ISO 8601 date formatters work as expected

brockboland:
See https://developer.apple.com/library/mac/qa/qa1480/_index.html

@vokal/ios-developers quick one

### DIFF
--- a/Pod/Classes/VOKManagedObjectMap.m
+++ b/Pod/Classes/VOKManagedObjectMap.m
@@ -8,6 +8,8 @@
 #import "VOKManagedObjectMap.h"
 #import "VOKCoreDataManager.h"
 
+NSString *const VOKDefaultDateFormatterLocaleIdentifier = @"en_US_POSIX";
+
 @implementation VOKManagedObjectMap
 
 + (instancetype)mapWithForeignKeyPath:(NSString *)inputKeyPath coreDataKey:(NSString *)coreDataKey
@@ -57,7 +59,7 @@
         DefaultDateFormatter = [NSDateFormatter new];
         DefaultDateFormatter.dateFormat = @"yyyy-MM-dd'T'HH:mm:ss.SSSSSS'Z'";
         DefaultDateFormatter.timeZone = [NSTimeZone timeZoneForSecondsFromGMT:0];
-        DefaultDateFormatter.locale = [NSLocale localeWithLocaleIdentifier:@"en_US_POSIX"];
+        DefaultDateFormatter.locale = [NSLocale localeWithLocaleIdentifier:VOKDefaultDateFormatterLocaleIdentifier];
     });
 
     return DefaultDateFormatter;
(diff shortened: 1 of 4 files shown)


chillpop:
LGTM, do you think it's worth adding a test for this?

brockboland:
Probably, but don't have time to do so right now. Feel like taking a crack at it?

chillpop:
sure

brockboland:
@chillpop's test has been merged. Anyone else want to have a look, @vokal/ios-developers?

brockboland:
Oh wait, one more thing: I'll do a bugfix release for this.

brockboland:
I take that back: updating the version number in the podspec means cascading updates to the pods directory in the sample projects, so I'll open another PR for that after this one is merged.

Pod/Classes/VOKManagedObjectMap.m
```diff
@@ -57,6 +57,7 @@ + (NSDateFormatter *)vok_defaultDateFormatter
         DefaultDateFormatter = [NSDateFormatter new];
         DefaultDateFormatter.dateFormat = @"yyyy-MM-dd'T'HH:mm:ss.SSSSSS'Z'";
         DefaultDateFormatter.timeZone = [NSTimeZone timeZoneForSecondsFromGMT:0];
+        DefaultDateFormatter.locale = [NSLocale localeWithLocaleIdentifier:@"en_US_POSIX"];
```
bryanluby: Would it make sense to make a constant out of this string literal in the .h file? That way it can be shared here and in tests?

seanwolter:
Neato. 
https://developer.apple.com/library/ios/qa/qa1480/_index.html

bryanluby:
@brockboland This would make a great show & tell at the next NSCoder

brockboland:
@bryanluby good points

brockboland:
Updated

bryanluby:
LGTM

SampleProject/VOKCoreDataManagerTests/DateFormatterTests.m
```diff
@@ -0,0 +1,27 @@
...
+
+@implementation DateFormatterTests
+
+- (void)testDateFormattersLocaleIsPOSIX
+{
+    XCTAssertEqualObjects([VOKManagedObjectMap vok_defaultDateFormatter].locale.localeIdentifier,
+                          VOKDefaultDateFormatterLocaleIdentifier);
+    XCTAssertEqualObjects([VOKManagedObjectMap vok_dateFormatterWithoutMicroseconds].locale.localeIdentifier,
+                          VOKDefaultDateFormatterLocaleIdentifier);
+}
```
vokal-isaac: My instinct is that it'd be better to test the output of the date formatter when applied to various date strings with various current locales, but I'm not sure offhand how to set the current locale programmatically for testing.

chillpop: AFAIK, you can only set the locale by passing in process arguments or choosing a locale in the Xcode scheme:
https://developer.apple.com/library/ios/documentation/MacOSX/Conceptual/BPInternational/TestingYourInternationalApp/TestingYourInternationalApp.html
http://useyourloaf.com/blog/using-launch-arguments-to-test-localizations/
Which makes testing that way a bit of a pain.

brockboland: Unless we move this test to a new test target and add a scheme specifically for it. I haven't done any testing to figure out what region or locale we could use for this that would change the calendar. You can switch it in Settings > General > Language & Region > Calendar.

vokal-isaac: Sounds like more trouble than it's worth, then.

brockboland:
@vokal-isaac good to pull, then?

vokal-isaac:
(:goat:,)

brockboland:
thanks!

PR #83, opened 2016-04-30
Remove default_subspecs on subspec

brockboland:
See https://github.com/CocoaPods/CocoaPods/issues/5228 for context. We can no longer include `default_subspecs` on a subspec, but the good news is, we don't need to. I tested that just adding `Vokoder/DataSources` as a pod will indeed include all the subspecs, which is what it was doing explicitly before anyway.

Most of the changes in here are just updates to the sample projects with the changed podspec.

@vokal/ios-developers can I get a review?


chillpop:
LGTM, not sure how I was able to get this working for the previous version 😕

brockboland:
¯\_(ツ)_/¯

PR #90, opened 2016-09-22
Swift 3 update

brockboland:
Unfortunately, this includes a lot of pod update garbage in the Swift sample project. Sorry about that.

@vokal/ios-developers can I get some review please?

PR #91, opened 2016-09-23
Use fastlane to run tests

brockboland:


brockboland:
@vokal/ios-developers moved things around a little bit here. As noted in #90, `xctool` doesn't support Xcode 8 yet, and we're using fastlane for all of our other projects anyway, so this switch made sense.

These sample projects are using a stripped-down version of our standard Fastfile, since there's no need for distribution, tagging on merge, etc. It's using a different matrix configuration to run tests in the two test projects in parallel. Note that the `Paging Data Source Example` project is no longer included, since it doesn't include any tests, which causes `scan` to fail.

vokal-isaac:
If you add an empty test target, it'll make `scan` happy and then at least fail if compilation fails.

brockboland:
I thought it had one already, but I guess I didn't actually check before.

PR #80, opened 2016-03-09
Add tvOS and OS X example targets to run tests

chillpop:
@vokal/ios-developers, @seanwolter, @designatednerd, code review?

This is almost entirely new target files and the results of running `pod install`.  There are relevant changes to the travis file at the top of the diff.

PR #92, opened 2016-09-26
Allow setting an explicit database URL

chillpop:
Before this change, the persistent store was automatically created in the app's Library folder.  This change allows users to explicitly set the file URL where the persistent store is created.  To allow for sharing a persistent store between an app and an extension, the persistent store should be in the app group's shared container.

@brockboland, @vokal-isaac, @bryanluby, @designatednerd code review?

NOTE: this is a pull request against the `swift2` branch.  I'd request that a release be made from that branch before merging back into `master`.


SampleProject/VOKCoreDataManagerTests/DatabaseURLTests.m
```diff
@@ -0,0 +1,74 @@
...
+    [super setUp];
+    
+    [[VOKCoreDataManager sharedInstance] resetCoreData];
+}
+
++ (void)tearDown
+{
+    [super tearDown];
+    
+    [[VOKCoreDataManager sharedInstance] resetCoreData];
```
brockboland: `[super teardown]` should be called after this test's teardown logic

brockboland:
@bryanluby @vokal-isaac either of you want to take a look before this is merged?

SampleProject/VOKCoreDataManagerTests/DatabaseURLTests.m
```diff
@@ -0,0 +1,74 @@
...
+}
+
+- (void)testInMemoryDatabaseURL
+{
+    //with an in-memory store, the persistentStoreFileURL should be nil
+    [[VOKCoreDataManager sharedInstance] setResource:@"VOKCoreDataModel" database:nil];
+    
+    XCTAssertNil([VOKCoreDataManager sharedInstance].persistentStoreFileURL);
+}
+
```
designatednerd: worth throwing in a test here where it creates the store in the test bundle to make sure that creating a store in a passed-in bundle works?

chillpop:
The Travis failure is on the Swift subspec and appears to have been happening since https://github.com/vokal/Vokoder/commit/8fd31ae81155a1cb9ab1aa09a320c54557f1bf39
Any thoughts @brockboland?

brockboland:
I'm not entirely sure what I did wrong here. I had updated the sample project to use CocoaPods 1.x while running Xcode 8, and thought I had the syntax right—it works in Xcode 8, anyway—but maybe not.

I switched back to 7 and fiddled with this a bit, and found things worked again after I used `pod init` to see what CocoaPods thought the structure of the Podfile should be. This is only a minor tweak from what it was, but it fixes the problems with the testing target. Could you try using this Podfile and see if things work locally for you?

```ruby
target 'SwiftyVokoder' do
  use_frameworks!

  pod 'Vokoder/Swift', :path => '../'
  pod 'xUnique', :git => 'https://github.com/vokal/xUnique.git', :commit => '43d39de'

  target 'SwiftyVokoderTests' do
    inherit! :search_paths
  end
end
```

chillpop:
I was able to get the Swift example to build locally after re-running `pod install` locally.  I can make a commit that has all of the sample projects updated with `pod install` but the diff may be huge.

There seems to be another issue with `pod lib lint` in the `after_script` step: it is failing trying to validate the MapperMacros subspec (because it's just a header file).  This is failing on Travis but not failing the build.

brockboland:
For the `swift2` branch, that's fine by me. The core changes made here will need to be manually cherry-picked into a new PR for master, but I don't foresee many additional feature additions or bug fixes happening to this branch, so maintaining parity between `master` and `swift2` shouldn't become an issue.

@bryanluby @vokal-isaac any thoughts on this?

brockboland:
Oh, as for `pod lib lint`: I wonder if that's due to the version of cocoapods that's running on Travis. `swift2` was branched from a commit before the Gemfile was added, and the [Travis docs](https://docs.travis-ci.com/user/osx-ci-environment/#Gems-in-the-global-gem-set) don't specify what version is installed. According to [CocoaPod docs](https://guides.cocoapods.org/making/using-pod-lib-create.html#deploying-your-library) (emphasis mine):

> The difference between them is that **`pod lib lint` does not access the network**, whereas `pod spec lint` checks the external repo and associated tag.

I would assume, then, that running `pod lib lint` would fail if Travis is still running 0.39. Is it passing locally?

chillpop:
No, `pod lib lint` fails locally too.  Because it doesn't fail the build, I'm guessing it's been failing since we added that subspec but we never noticed.

chillpop:
...or since we added `pod lib lint` to the travis file: https://github.com/vokal/Vokoder/pull/80

chillpop:
I hate Travis

brockboland:
It works locally?

chillpop:
Not working locally.  Trying to nuke the pods and reinstall.

chillpop:
I think I've got it.  The nested target pod framework naming scheme changed between 0.39 and 1.0 but the old framework was still referenced in the project file.

chillpop:
GREEEEEEEEEEEEEEEN!

PR #95, opened 2016-10-04
Mark VOKMigrationFailureOptionWipeRecoveryAndAlert deprecated

brockboland:
@chillpop @vokal/ios-developers cool?


brockboland:
Note that this is against the `swift2` branch

PR #101, opened 2016-10-13
Allow unshared instances of the manager

vokal-isaac:
In order to allow multiple instances of `VOKCoreDataManager` to coexist:
- any instance methods on `VOKCoreDataManager` should use `self` instead of the shared instance
- any class methods that require an instance of `VOKCoreDataManager` should be converted to instance methods (***breaking change***)
- `VOKCoreDataManager` should not use any methods from `NSManagedObject+VOKManagedObjectAdditions` that require a manager instance

The methods in `NSManagedObject+VOKManagedObjectAdditions` that require a manager instance are now all convenience wrappers that call instance methods on the shared manager.  This means that having multiple manager instances means using the manager methods directly rather than the convenience category methods on the managed objects.

@vokal/ios-developers @chillpop Code review please?

### DIFF
--- a/Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
+++ b/Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
@@ -34,48 +34,7 @@
 
 + (NSString *)vok_entityName
 {
-    static char AssociatedObjectKey;  // The key for the runtime-association.
-    
-    // Get the associated value.
-    NSString *vok_entityName = objc_getAssociatedObject(self, &AssociatedObjectKey);
-    
-    // If we didn't find an associated entity name, determine the entity name.
-    if (!vok_entityName) {
-        // If we already have an entityName class method (e.g., MOGenerator-generated subclasses), use it.
-        // (Note that we have to cast self (a Class) to id to use NSObject's dynamic-selector methods, even though they work.)
-        if ([(id)self respondsToSelector:@selector(entityName)]) {
-            vok_entityName = [(id)self performSelector:@selector(entityName)];
-        }
-        if (!vok_entityName) {
-            // On OS X, NSObject has a private class method called entityName but it may return nil.
-            // https://github.com/rentzsch/mogenerator/issues/196
-            
-            // Since we don't have an entityName class method (or it didn't return a result),
-            // look up the entity name in the managed object model.
-            NSManagedObjectModel *model = [[VOKCoreDataManager sharedInstance] managedObjectModel];
-            
-            // Start with the class we are...
-            Class workingClass = self;
-            do {
-                NSString *workingClassName = NSStringFromClass(workingClass);
-                // ... check for a matching entity in the model...
-                for (NSEntityDescription *description in model.entities) {
-                    if ([workingClassName isEqualToString:description.managedObjectClassName]) {
-                        vok_entityName = description.name;
-                        break;
-                    }
-                }
-                // ... and walk up the superclass chain...
-                workingClass = [workingClass superclass];
-                // ... until we get Nil or find a matching entity (as long as we have a superclass to test and haven't found the entity name).
-            } while (workingClass && !vok_entityName);
-        }
-        NSAssert(vok_entityName, @"no entity found that uses %@ as its class", NSStringFromClass(self));
-        // Save the determined entity name as an associated value.
-        objc_setAssociatedObject(self, &AssociatedObjectKey, vok_entityName, OBJC_ASSOCIATION_RETAIN_NONATOMIC);
-    }
-    
-    return vok_entityName;
+    return [[VOKCoreDataManager sharedInstance] entityNameForClass:self];
 }
 
 #pragma mark - Create Objects
--- a/Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
+++ b/Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
@@ -7,8 +7,6 @@
 
 #import "NSManagedObject+VOKManagedObjectAdditions.h"
 
-#import <objc/runtime.h>
-
 #import "VOKCoreDataManager.h"
 
 @implementation NSManagedObject (VOKManagedObjectAdditions)
--- a/Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
+++ b/Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
@@ -113,19 +113,20 @@
 
 + (void)vok_addWithArrayInBackground:(NSArray *)inputArray completion:(VOKManagedObjectsReturnBlock)completion
 {
-    [VOKCoreDataManager importArrayInBackground:inputArray
-                                       forClass:[self class]
-                                     completion:^(NSArray *arrayOfManagedObjectIDs) {
-                                         if (completion) {
-                                             // if there is no completion block there's no need to collect the new/updated objects
-                                             NSMutableArray *returnArray = [NSMutableArray arrayWithCapacity:arrayOfManagedObjectIDs.count];
-                                             NSManagedObjectContext *moc = [[VOKCoreDataManager sharedInstance] managedObjectContext];
-                                             for (NSManagedObjectID *objectID in arrayOfManagedObjectIDs) {
-                                                 [returnArray addObject:[moc objectWithID:objectID]];
-                                             }
-                                             completion([returnArray copy]);
-                                         }
-                                     }];
+    VOKCoreDataManager *coreDataManager = VOKCoreDataManager.sharedInstance;
+    [coreDataManager importArrayInBackground:inputArray
+                                    forClass:[self class]
+                                  completion:^(NSArray *arrayOfManagedObjectIDs) {
+                                      if (completion) {
+                                          // if there is no completion block there's no need to collect the new/updated objects
+                                          NSMutableArray *returnArray = [NSMutableArray arrayWithCapacity:arrayOfManagedObjectIDs.count];
+                                          NSManagedObjectContext *moc = [coreDataManager managedObjectContext];
+                                          for (NSManagedObjectID *objectID in arrayOfManagedObjectIDs) {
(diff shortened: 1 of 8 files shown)


chillpop:
Is the idea to allow multiple instances in addition to the sharedInstance, or deprecate the sharedInstance entirely?

vokal-isaac:
In addition to `sharedInstance`.  I think the vast majority of the time, `sharedInstance` is the way to go.  I just happen to be looking at a particular use case on macOS where the data for a document is represented with an in-memory Core Data store and multiple documents can be open at once.  I think having a manager per document is cleaner than designing the Core Data models to allow multiple documents' models to coexist in a single store.

Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
```diff
@@ -113,19 +72,20 @@ + (instancetype)vok_addWithDictionary:(NSDictionary *)inputDict forManagedObject
...
-                                             NSMutableArray *returnArray = [NSMutableArray arrayWithCapacity:arrayOfManagedObjectIDs.count];
-                                             NSManagedObjectContext *moc = [[VOKCoreDataManager sharedInstance] managedObjectContext];
-                                             for (NSManagedObjectID *objectID in arrayOfManagedObjectIDs) {
-                                                 [returnArray addObject:[moc objectWithID:objectID]];
-                                             }
-                                             completion([returnArray copy]);
-                                         }
-                                     }];
+    VOKCoreDataManager *coreDataManager = VOKCoreDataManager.sharedInstance;
+    [coreDataManager importArrayInBackground:inputArray
```
chillpop: What's the plan for using this and other convenience methods with a non-shared VOKCoreDataManager?  You can't?

Pod/Classes/Optional Data Sources/VOKFetchedResultsDataSource.h
```diff
@@ -224,6 +224,33 @@ NS_ASSUME_NONNULL_BEGIN
...
+ @param delegate           Delegate to notify of object selection/deselection and whether the fetch has results
+ 
+ @return New data source
+ */
+- (instancetype)initWithCoreDataManager:(VOKCoreDataManager *)coreDataManager
+                              predicate:(NSPredicate *)predicate
+                              cacheName:(NSString *)cacheName
+                              tableView:(UITableView *)tableView
+                     sectionNameKeyPath:(NSString *)sectionNameKeyPath
+                        sortDescriptors:(NSArray *)sortDescriptors
```
chillpop: I believe all the parameters from here and above should be marked `nullable`.
Also: this one should be `VOKArrayOfSortDescriptors` to get a typed array

Pod/Classes/Optional Data Sources/VOKFetchedResultsDataSource.h
```diff
@@ -224,6 +224,33 @@ NS_ASSUME_NONNULL_BEGIN
...
+                     managedObjectClass:(Class)managedObjectClass
+                              batchSize:(NSInteger)batchSize
+                             fetchLimit:(NSInteger)fetchLimit
+                               delegate:(id <VOKFetchedResultsDataSourceDelegate>)delegate;
```
chillpop: nullable here too

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -97,7 +99,7 @@ - (void)setResource:(nullable NSString *)resource
     }
     
     // Touch the managed object context to ensure it's been created
-    [[VOKCoreDataManager sharedInstance] managedObjectContext];
+    [self managedObjectContext];
```
chillpop: `self.managedObjectContext`?

Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
vokal-isaac: Right.  Basically, the convenience methods in `NSManagedObject+VOKManagedObjectAdditions` are (now that `vok_entityName` has been rewritten) just wrappers around methods on the manager, so if one needs a manager-specific call, passing the manager in as an argument to a method on the managed object seemed uglier than calling the method on the manager with the managed object as an argument.

There probably should be an explicit note to the effect of "things in `NSManagedObject+VOKManagedObjectAdditions` use `VOKCoreDataManager.sharedInstance`."

Pod/Classes/Optional Data Sources/VOKFetchedResultsDataSource.h
vokal-isaac: Whoops.  I really wish Xcode's autocomplete on the `.m` side would carry over the nullability annotations, etc., from the `.h` side.  (This was copied back from the `.m`, hence missing all that stuff.)

Pod/Classes/VOKCoreDataManager.m
chillpop: Enh, keep it as is.  I wasn't paying attention to the context of the call.

Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
chillpop: Sound good.

Pod/Classes/NSManagedObject+VOKManagedObjectAdditions.m
```diff
@@ -34,48 +34,7 @@ - (NSDictionary *)vok_dictionaryRepresentationRespectingKeyPaths
...
-                // ... until we get Nil or find a matching entity (as long as we have a superclass to test and haven't found the entity name).
-            } while (workingClass && !vok_entityName);
-        }
-        NSAssert(vok_entityName, @"no entity found that uses %@ as its class", NSStringFromClass(self));
-        // Save the determined entity name as an associated value.
-        objc_setAssociatedObject(self, &AssociatedObjectKey, vok_entityName, OBJC_ASSOCIATION_RETAIN_NONATOMIC);
-    }
-    
-    return vok_entityName;
+    return [[VOKCoreDataManager sharedInstance] entityNameForClass:self];
```
vokal-isaac: My inclination was to leave it in place in the same spirit as the rest of the methods in the category—they're fine if you're only using `sharedInstance` and they're totally wrong/broken if you're not.

brockboland: That expectation should be prominently documented in the header for this category.

Pod/Classes/Optional Data Sources/VOKFetchedResultsDataSource.m
```diff
@@ -59,6 +63,28 @@ - (instancetype)initWithPredicate:(NSPredicate *)predicate
                   sortDescriptors:(NSArray *)sortDescriptors
                managedObjectClass:(Class)managedObjectClass
                         batchSize:(NSInteger)batchSize
+                       fetchLimit:(NSInteger)fetchLimit
+                         delegate:(id <VOKFetchedResultsDataSourceDelegate>)delegate
+{
+    return [self initWithCoreDataManager:VOKCoreDataManager.sharedInstance
```
brockboland: I would pass `nil` here to let the designated initializer handle it

Pod/Classes/VOKCoreDataManager.h
```diff
@@ -302,7 +309,7 @@ typedef void(^VOKObjectIDsReturnBlock)(VOKArrayOfManagedObjectIDs *managedObject
  @param objectClass     Specifies the class to instantiate or fetch when importing data.
  @param completion      Fired on the main queue once the changes have been merged. It brings an NSArray of permanent NSManagedObjectIDs matching the objects deserialized from the import array.
  */
-+ (void)importArrayInBackground:(VOKArrayOfObjectDictionaries *)inputArray
+- (void)importArrayInBackground:(VOKArrayOfObjectDictionaries *)inputArray
```
brockboland: I can't find it now, but I thought there was a compiler macro to indicate that some method has been replaced with another one, so that Xcode can suggest the appropriate correction. I can't find it now, but if that rings any bells, it might be helpful to include it here.

brockboland: …here and for `writeToTemporaryContext` above, that is.

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -8,6 +8,8 @@
 #import "VOKCoreDataManager.h"
 #import "VOKCoreDataManagerInternalMacros.h"
 
+#import <objc/runtime.h>
```
brockboland: Looks like this import can be removed from `NSManagedObject+VOKManagedObjectAdditions.m`. I can't comment on the line over there because it's not part of the diff.

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -631,36 +682,38 @@ - (void)saveAndMergeWithMainContextAndWait:(NSManagedObjectContext *)context;
 
 #pragma mark - Background Importing
 
-+ (void)writeToTemporaryContext:(VOKWriteBlock)writeBlock
+- (void)writeToTemporaryContext:(VOKWriteBlock)writeBlock
```
brockboland: Same thing in a couple places here, regarding that compiler macro for replaced methods (if we find it)

Pod/Classes/VOKCoreDataManager.h
vokal-isaac: If I left the old methods in place, they could be tagged with the deprecation macro with a message pointing to the new methods.

I'd say maybe we add the instance methods to 3.x and mark the class methods as deprecated there, with the note that they'll be removed in 4.x?

brockboland: We discussed this in person, but: nah, I'd rather just move forward with the changes in 4.x and leave 3.x where it is.

brockboland:
@chillpop @bryanluby any other feedback on this one?

chillpop:
still LGTM

bryanluby:
LGTM

PR #104, opened 2016-10-26
Removed VOKMigrationFailureOptionWipeRecoveryAndAlert

brockboland:
- Removed VOKMigrationFailureOptionWipeRecoveryAndAlert
- Added VOKMigrationFailureWipeRecoveryNotificationName
- Added a new project to test this
- Updated the CocoaPods version used

This one is big because of the new project, but these are the files that actually matter:

- `.travis.yml`
- `Gemfile`
- `MigrationProject/MigrationProjectTests/MigrationProjectTests.m`
- `Pod/Classes/VOKCoreDataManager.[m|h]`

@vokal/ios-developers review please?


brockboland:
cc @designatednerd @chillpop

brockboland:
Fixes #96

MigrationProject/MigrationProject/AppDelegate.m
```diff
@@ -0,0 +1,18 @@
...
+//  Created by Brock Boland on 10/26/16.
+//  Copyright © 2016 Vokal. All rights reserved.
+//
+
+#import "AppDelegate.h"
+
+// This app does nothing. This project is only here for tests.
+@implementation AppDelegate
+
+- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
```
designatednerd: i think you can even delete this method too

MigrationProject/MigrationProjectTests/MigrationProjectTests.m
```diff
@@ -0,0 +1,71 @@
...
+
+    [[VOKCoreDataManager sharedInstance] resetCoreData];
+
+    // Load a copy of the DB that used the first version of the model
+    NSBundle *bundle = [NSBundle bundleForClass:[self class]];
+    NSURL *oldPersistentStoreURL = [bundle URLForResource:@"MyDB" withExtension:@"sqlite"];
+
+    XCTAssertNotNil(oldPersistentStoreURL, @"Cannot find MyDB.sqlite");
+
+    NSURL *docsDir = [[NSFileManager defaultManager] URLsForDirectory:NSDocumentDirectory inDomains:NSUserDomainMask].lastObject;
```
designatednerd: worth asserting that docs dir is non-nil?

MigrationProject/MigrationProjectTests/MigrationProjectTests.m
```diff
@@ -0,0 +1,71 @@
...
+    self.persistentStoreURL = [docsDir URLByAppendingPathComponent:@"testdb"];
+    
+    // Copy the old version of the database into the documents directory to work with
+    NSError *error;
+     [[NSFileManager defaultManager] copyItemAtURL:oldPersistentStoreURL
+                                             toURL:self.persistentStoreURL
+                                             error:&error];
+     XCTAssertNil(error);
```
designatednerd: in b4 @vokal-isaac talks about how you should also check the result of this and not just the error - a nil error is not a guarantee of success

MigrationProject/MigrationProjectTests/MigrationProjectTests.m
```diff
@@ -0,0 +1,71 @@
...
+}
+
+- (void)tearDown
+{
+    // Make sure the file doesn't lingered around after the test is done
```
designatednerd: rm `ed` from lingered

designatednerd:
One question from me: Why does this need to be in a totally separate project from other stuff? It seems like you should be able to add that core data model to one of the existing projects and load it from there.

brockboland:
Because I need to force it to fail to migrate the store, I was thinking that it would need to be completely separate from other sample projects—but if I'm specifying the model AND the store URL for this test case anyway, it should be fine. Let me try it out, and I might open a new PR instead of polluting the git history with this project if it's not needed.

brockboland:
Moving to #105

PR #105, opened 2016-10-27
Removed VOKMigrationFailureOptionWipeRecoveryAndAlert, take 2

brockboland:
Simplified version of #104 

- Removed `VOKMigrationFailureOptionWipeRecoveryAndAlert`
- Added `VOKMigrationFailureWipeRecoveryNotificationName`
- Added a test for this
- Updated the CocoaPods version used
- Removed `VOK_TARGET_USES_UIKIT`

@vokal/ios-developers @chillpop apologies for this: @designatednerd was right, there was no need for a separate project to test this change.


brockboland:
Oh, @designatednerd, I also addressed your comments in the tests

Pod/Classes/VOKCoreDataManager.m
```diff
@@ -175,46 +177,18 @@ - (void)addPersistentStoreToCoordinator:(NSPersistentStoreCoordinator *)persiste
                                                         options:options
                                                           error:&error]) {
         switch (self.migrationFailureOptions) {
-            case VOKMigrationFailureOptionWipeRecoveryAndAlert:
-#if VOK_TARGET_USES_UIKIT
```
chillpop: I think this macro can be deleted since it was only used here

brockboland: Good point!

Pod/Classes/VOKCoreDataManager.h
```diff
@@ -25,10 +25,14 @@ typedef NS_ENUM (NSInteger, VOKMigrationFailureOption) {
     VOKMigrationFailureOptionNone,
     /// Silently delete and recreate the database file, data will be erased, but instability and crashing will be avoided
     VOKMigrationFailureOptionWipeRecovery,
-    /// Same as VOKMigrationFailureOptionWipeRecoveryAndAlert, but will also notify the user that data has been erased via an on screen alert. NOTE: not yet implemented on OS X
-    VOKMigrationFailureOptionWipeRecoveryAndAlert,
 };
 
+/**
+ * Name of the NSNotification that's posted after a successful DB wipe following a migration failure.
```
chillpop: If you're adding this notification on success, worth adding one for failure too?

brockboland: It already raises an `NSException` in that case, so I think we're good

chillpop: oh, I see, I was misreading the comment.  This is exactly what I was expecting

PR #106, opened 2016-10-27
Update CocoaPods in sample projects

brockboland:
All I did was touch up one of the Podfiles, then `bundle exec pod install` in all the sample project directories to get them up to 1.1.1. The tests all still pass locally and it made one hell of a diff.


brockboland:
I just restarted two of the five jobs. They're still occasionally failing with `Exit code 65` for no apparent reason.

brockboland:
Welp, `scan` update didn't fix it on the first try, but it worked on the second!

@bryanluby if you're OK with the `scan` update, could you go ahead and merge this in the morning?

bryanluby:
@brockboland sure thing

PR #110, opened 2016-11-08
Updated version requirements in README

brockboland:
Of course I noticed this right after pushing 4.0.1 to CocoaPods.

I dropped the bit about the oldest version of Xcode required (I'm sure it's no longer 5.1), and I'm not sure how far back the unit tests will run but don't want to keep making promises there. Instead, I specified earliest OS versions, based on the podspec.

@vokal/ios-developers review?

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ If you intend to use Vokoder in Swift, use the `Swift` subspec instead:
 
     pod "Vokoder/Swift"
 
-Vokoder requires Xcode 5.1 or higher.  The unit tests require features found in Xcode 6 and higher.  The Swift extensions require Swift 2 and Xcode 7.
+Vokoder requires iOS 7, tvOS 9, or macOS 10.9. The Swift extensions require Swift 3 and Xcode 8.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,7 +20,7 @@ If you intend to use Vokoder in Swift, use the `Swift` subspec instead:
 
     pod "Vokoder/Swift"
 
-Vokoder requires Xcode 5.1 or higher.  The unit tests require features found in Xcode 6 and higher.  The Swift extensions require Swift 2 and Xcode 7.
+Vokoder requires iOS 7, tvOS 9, or macOS 10.9. The Swift extensions require Swift 3 and Xcode 8.
```
vokal-isaac: Is it necessary to specify Xcode 8?

brockboland: Not really, that part I just updated. Doesn't really matter.

vokal-isaac: I'd be inclined to strip it out to avoid having to update it again when Xcode 9 ships, but :shrug:.

PR #111, opened 2016-11-10
Pod udpates

brockboland:
I didn't `pod install` in the sample projects after updating VOKUtilities in the podspec. This updates that, and preps the podspec here for the next release so we don't forget to increment the bugfix version.

@vokal/ios-developers quick review, please?


vokal-isaac:
On it.

PR #112, opened 2016-11-10
Update CocoaPods local repo before linting

brockboland:
And only do so on success. Hopefully this fixes the lint failure on Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,8 @@ env:
     - SCAN_WORKSPACE="SwiftSampleProject/SwiftSampleProject.xcworkspace"                 SCAN_SCHEME="SwiftyVokoder"              SCAN_SDK="iphonesimulator"
     - SCAN_WORKSPACE="Paging Data Source Example/Paging Data Source Example.xcworkspace" SCAN_SCHEME="Paging Data Source Example" SCAN_SDK="iphonesimulator"
 script: bundle exec scan
-after_script:
+after_success:
+  - bundle exec pod repo update # Travis isn't finding the latest VOKUtilities release: must have an old repo cached
   - bundle exec pod lib lint
 git:
   depth: 10000  # For the auto-incrementing build number script to work, we need to clone with a basically-infinite depth.


.travis.yml
```diff
@@ -11,7 +11,8 @@ env:
     - SCAN_WORKSPACE="SwiftSampleProject/SwiftSampleProject.xcworkspace"                 SCAN_SCHEME="SwiftyVokoder"              SCAN_SDK="iphonesimulator"
     - SCAN_WORKSPACE="Paging Data Source Example/Paging Data Source Example.xcworkspace" SCAN_SCHEME="Paging Data Source Example" SCAN_SDK="iphonesimulator"
 script: bundle exec scan
-after_script:
+after_success:
+  - bundle exec pod repo update # Travis isn't finding the latest VOKUtilities release: must have an old repo cached
```
vokal-isaac: Doesn't this have the potential to be really slow?

brockboland: From the error message that prompted this change:

> Note: as of CocoaPods 1.0, `pod repo update` does not happen on `pod install` by default.) during validation.

So: probably, but it was part of `pod install` in the past, so it can't be THAT bad.

vokal-isaac: I don't think we were running `pod install` in the past.

brockboland: Once the build finishes, I'll compare the run time against the last PR build: https://travis-ci.org/vokal/Vokoder/builds/174855777

brockboland: We weren't. I'm just saying, it can't be THAT bad, because we run `pod install` locally all the time and I don't remember too many gripes about how long that was taking. This command took 9 seconds on my machine just now, so we'll see what the impact is on Travis.

brockboland:
Oof: impact on runtime is considerable. Average run time for the five sub-builds increased more than 4x from 183s to 843s. Looking into options here.

brockboland:
I'm putting this one on hold for now, and closing without merging. The needed change here is being tracked in #103, and we'll revisit after Travis has an Xcode 8.1 image.

PR #113, opened 2016-11-11
Clean up some analyzer warnings and mark a method that can return nil

### LINKED ISSUE #107: Fix Analyzer issues
Using Xcode's "Analyze" shows several issues (I'm seeing 9 in the sample project and 2 in another app using Vokoder, with one one of those being common between the two projects).  We should fix those and, ideally, find a way to make analyzer warnings fail the build.

brockboland:
Mostly just nil checks.

Fixes #107.

@vokal/ios-developers cool?

### DIFF
--- a/Pod/Classes/Optional Data Sources/VOKCollectionDataSource.m
+++ b/Pod/Classes/Optional Data Sources/VOKCollectionDataSource.m
@@ -243,6 +243,11 @@
                         indexPath = obj[0];
                         newIndexPath = obj[1];
                     }
+                    
+                    // Can't do anything if we don't have an index path
+                    if (indexPath == nil) {
+                        return;
+                    }
 
                     // NSFetchedResultsController doesn't handle fetchLimit
                     // properly, so we need to check index paths against the
@@ -270,7 +275,9 @@
                             if (fetchLimit > 0) {
                                 if (indexPath.row < fetchLimit && newIndexPath.row < fetchLimit) {
                                     // Before and after are both in range
-                                    [self.collectionView moveItemAtIndexPath:indexPath toIndexPath:newIndexPath];
+                                    if (newIndexPath) {
+                                        [self.collectionView moveItemAtIndexPath:indexPath toIndexPath:newIndexPath];
+                                    }
                                 } else if (indexPath.row >= fetchLimit && newIndexPath.row >= fetchLimit) {
                                     // Both out of range: do nothing
                                 } else if (indexPath.row < fetchLimit && newIndexPath.row >= fetchLimit) {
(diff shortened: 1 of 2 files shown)


Pod/Classes/Optional Data Sources/VOKCollectionDataSource.m
```diff
@@ -270,19 +275,25 @@ - (void)controllerDidChangeContent:(NSFetchedResultsController *)controller
                             if (fetchLimit > 0) {
                                 if (indexPath.row < fetchLimit && newIndexPath.row < fetchLimit) {
                                     // Before and after are both in range
-                                    [self.collectionView moveItemAtIndexPath:indexPath toIndexPath:newIndexPath];
+                                    if (newIndexPath) {
+                                        [self.collectionView moveItemAtIndexPath:indexPath toIndexPath:newIndexPath];
+                                    }
```
vokal-isaac: Would it make sense to test `newIndexPath` at the top of the `NSFetchedResultsChangeMove` case?  It's being used in various tests outside of these checks as it is.

brockboland: Yeah, you're right. I thought it would still need to execute the case on line 285, but that's not true after all.

Pod/Classes/Optional Data Sources/VOKCollectionDataSource.m
```diff
@@ -272,28 +272,22 @@ - (void)controllerDidChangeContent:(NSFetchedResultsController *)controller
...
-                                        [self.collectionView insertItemsAtIndexPaths:@[newIndexPath]];
-                                    }
+                                    [self.collectionView insertItemsAtIndexPaths:@[newIndexPath]];
                                 }
                             } else {
                                 // No fetch limit: behave normally
-                                if (newIndexPath) {
-                                    [self.collectionView moveItemAtIndexPath:indexPath toIndexPath:newIndexPath];
-                                }
+                                [self.collectionView moveItemAtIndexPath:indexPath toIndexPath:newIndexPath];
```
vokal-isaac: Should this case execute if `newIndexPath` is `nil`?

Pod/Classes/Optional Data Sources/VOKCollectionDataSource.m
```diff
@@ -354,28 +348,22 @@ - (void)controllerDidChangeContent:(NSFetchedResultsController *)controller
...
-                                            [self.collectionView insertItemsAtIndexPaths:@[newIndexPath]];
-                                        }
+                                        [self.collectionView insertItemsAtIndexPaths:@[newIndexPath]];
                                     }
                                 } else {
                                     // No fetch limit: behave normally
-                                    if (newIndexPath) {
-                                        [self.collectionView moveItemAtIndexPath:indexPath toIndexPath:newIndexPath];
-                                    }
+                                    [self.collectionView moveItemAtIndexPath:indexPath toIndexPath:newIndexPath];
```
vokal-isaac: Ditto.

PR #121, opened 2017-01-23
Try pod lib lint

brockboland:
DO NOT MERGE

I'm just confirming this works. I'll open a new PR with just the relevant commit once it's ready.


brockboland:
My main concern with this was added build time, but the times on [this build](https://travis-ci.org/vokal/Vokoder/builds/194543196) are actually faster the [the build for my last PR](https://travis-ci.org/vokal/Vokoder/builds/193478723). Obviously, this change does't speed things up—it's surely just because Travis was busier last week—but this puts to rest my concern that it would have a huge negative impact on build times.

vokal-isaac:
`pod lib lint` added [a little over 3 minutes](https://travis-ci.org/vokal/Vokoder/jobs/194543201#L646) on one of those builds.
—that seems like a worthwhile tradeoff to me.

brockboland:
How had I never noticed that run time at the right there?

brockboland:
See #123

PR #125, opened 2017-01-24
More Travis matrix testing

brockboland:


brockboland:
Theory confirmed. Anything listed under `include` is added one-off, not multiplied against the existing matrix.

PR #126, opened 2017-01-24
One more Travis matrix test

brockboland:


brockboland:
Confirmed: Travis doesn't know what to make of this. The "Job log" tab doens't show the build jobs, and the "View config" tab indicates that it's ignoring the `matrix` key.

screen shot 2017-01-24 at 1 38 01 pm

.travis.yml
```diff
@@ -1,15 +1,15 @@
 language: objective-c
-osx_image: xcode8.2
```
vokal-isaac: As an alternate, leave everything in the `env` section as-is and change this to
```yaml
osx_image:
  matrix:
    - xcode8.1
    - xcode8.2
```

PR #124, opened 2017-01-24
Run Travis build on multiple Xcode versions

brockboland:
Another test, don't merge

PR #127, opened 2017-02-06
Run tests on multiple iOS versions

brockboland:
This is the culmination of all the testing and trial and error in #124

@vokal/ios-developers what do you think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,22 @@ env:
   global:
     - SCAN_SKIP_BUILD=1
   matrix:
-    # Define the details for each test suite that should be run
-    - SCAN_WORKSPACE="SampleProject/SampleProject.xcworkspace"                           SCAN_SCHEME="VOKCoreDataManager"         SCAN_SDK="iphonesimulator"
+    # Test on iOS 10.2
+    - SCAN_WORKSPACE="SampleProject/SampleProject.xcworkspace"                           SCAN_SCHEME="VOKCoreDataManager"         SCAN_DEVICE="iPhone 5 (10.2)"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,12 +4,22 @@ env:
   global:
     - SCAN_SKIP_BUILD=1
   matrix:
-    # Define the details for each test suite that should be run
-    - SCAN_WORKSPACE="SampleProject/SampleProject.xcworkspace"                           SCAN_SCHEME="VOKCoreDataManager"         SCAN_SDK="iphonesimulator"
+    # Test on iOS 10.2
+    - SCAN_WORKSPACE="SampleProject/SampleProject.xcworkspace"                           SCAN_SCHEME="VOKCoreDataManager"         SCAN_DEVICE="iPhone 5 (10.2)"
```
bryanluby: Is there reason for using iPhone 5 as the device rather than iPhone 6 or 6S

brockboland: Partly because fastlane defaults to the 5s if you don't specify (so I figured, that seems fine), and partly because it tends to be useful to test on older/smaller devices to catch edge cases that we might not see in our own testing.

On the other hand, I suppose that this reasoning should lead to the 5—but I also suspect that usage rates on that are low enough that it can be ignored. But then again, it's also the last 32-bit iPhone, so maybe it's worth testing there for that reason.

I guess I could go either way, on the 5 vs. 5s. What do you think?

bryanluby: Prob fine as is for now

brockboland: Oh right, I forgot that [32-bit support isn't really a thing](https://arstechnica.com/apple/2017/01/future-ios-release-will-soon-end-support-for-unmaintained-32-bit-apps/?utm_campaign=iOS%2BDev%2BWeekly&utm_medium=email&utm_source=iOS_Dev_Weekly_Issue_286) anymore anyway.

PR #128, opened 2017-03-17
Add watchOS support, drop iOS 7 support

designatednerd:
This PR adds watchOS 3.0 support to Vokoder, and drops iOS 7 support since it has been dropped by a) VOKUtilities and b) Apple. 

Also updated the bundle file to use the release version of CocoaPods 1.2.0.


designatednerd:
@vokal-isaac @bryanluby @lindelldigital @brockboland @chillpop Please rip this to shreds. Thank you!

Paging Data Source Example/Paging Data Source Example.xcodeproj/project.pbxproj
```diff
@@ -523,40 +529,38 @@
...
 				DEVELOPMENT_TEAM = TZPBQT6RZA;
-				ENABLE_NS_ASSERTIONS = NO;
 				GCC_C_LANGUAGE_STANDARD = gnu99;
 				GCC_WARN_ABOUT_RETURN_TYPE = YES_ERROR;
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				INFOPLIST_FILE = "Paging Data Source ExampleTests/Info.plist";
 				IPHONEOS_DEPLOYMENT_TARGET = 10.0;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
-				MTL_ENABLE_DEBUG_INFO = NO;
+				MTL_ENABLE_DEBUG_INFO = YES;
```
chillpop: :metal:

Vokoder.podspec
```diff
@@ -1,15 +1,16 @@
 Pod::Spec.new do |s|
   s.name             = "Vokoder"
-  s.version          = "4.1.1"
+  s.version          = "5.0.0"
```
chillpop: I would argue for 4.2.0
This isn't a breaking change for the existing platforms, yes?

chillpop:
Should there be an update to the `.travis.yml` to test on watchOS?

Vokoder.podspec
designatednerd: It drops iOS 7.0 support, so yes, I would argue that it is.

designatednerd:
@chillpop Sigh. watchOS does not support testing. 

The main way this is being tested right now is by adding the watchOS app to the main app - that way if it doesn't compile or import, we'll hear right from there.

designatednerd:
the one which failed seems to run fine locally - anyone with access wanna hit the "go again" button?

designatednerd:
(thanks whoever mashed the button!)

designatednerd:
light green trap clean! thoughts?

designatednerd:
🎉 danke! plz give me a heads up when it's tagged and up on CP.

vokal-isaac:
Tagged.  I'm having issues pushing it to trunk, though.

designatednerd:
@vokal-isaac ugh, hold on a second if you haven't suceeded - looks like I need to make a couple updates to the podfile to make it work right with swift

PR #129, opened 2017-03-17
RM alternate dependency declarations

designatednerd:
Derp, still couldn't install the swift subspec on watchOS because that subspec had a different restriction because of former dependency on older version of iOS. 

Bumped the spec version but I can take it off if we want to re-tag master 5.0.0

### DIFF
--- a/Vokoder.podspec
+++ b/Vokoder.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = "Vokoder"
-  s.version          = "5.0.0"
+  s.version          = "5.0.1"
   s.summary          = "Vokal's Core Data Manager"
   s.homepage         = "https://github.com/vokal/Vokoder"
   s.license          = { :type => "MIT", :file => "LICENSE"}
@@ -32,8 +32,6 @@ Pod::Spec.new do |s|
 
   s.subspec 'DataSources' do |ss|
     ss.dependency 'Vokoder/Core'
-    ss.ios.deployment_target = '8.0'
-    ss.tvos.deployment_target = '9.0'
 
     ss.subspec 'FetchedResults' do |sss|
       sss.source_files = 'Pod/Classes/Optional Data Sources/VOKFetchedResultsDataSource.{h,m}'
@@ -54,8 +52,6 @@ Pod::Spec.new do |s|
   end
 
   s.subspec 'Swift' do |sw|
-    sw.ios.deployment_target = '8.0'
-    sw.tvos.deployment_target = '9.0'
     sw.dependency 'Vokoder/DataSources'
     sw.source_files = 'Pod/Classes/Swift/*.swift'
   end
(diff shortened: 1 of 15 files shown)


designatednerd:
@vokal-isaac @bryanluby @lindelldigital @chillpop @brockboland derp.

Vokoder.podspec
```diff
@@ -32,8 +32,6 @@ Pod::Spec.new do |s|
 
   s.subspec 'DataSources' do |ss|
     ss.dependency 'Vokoder/Core'
-    ss.ios.deployment_target = '8.0'
-    ss.tvos.deployment_target = '9.0'
```
chillpop: You may still need this since this won't compile on macOS due to UIKit stuff.

Vokoder.podspec
```diff
@@ -54,8 +52,6 @@ Pod::Spec.new do |s|
   end
 
   s.subspec 'Swift' do |sw|
-    sw.ios.deployment_target = '8.0'
-    sw.tvos.deployment_target = '9.0'
```
chillpop: Same thing here due to the dependency on the DataSources

Vokoder.podspec
chillpop: Would it be better to add watchOS explicitly here and below instead of removing everything?

designatednerd: oh goddamn it you're right. BAH BAH BAH.

designatednerd:
Woof. looks like i'm gonna have to rearrange some stuff since the swift podspec depends on the data sources podspec, which has a bunch of stuff that can't be used in watchOS.

designatednerd:
Aha, better solution: Swift subspec doesn't actually need to depend on anything outside of `Core`, so I just switched the dependency.

Vokoder.podspec
```diff
@@ -56,7 +56,9 @@ Pod::Spec.new do |s|
   s.subspec 'Swift' do |sw|
     sw.ios.deployment_target = '8.0'
     sw.tvos.deployment_target = '9.0'
-    sw.dependency 'Vokoder/DataSources'
+    sw.watchos.deployment_target = '3.0'
+    
+    sw.dependency 'Vokoder/Core'
```
chillpop: The advantage of (and intention) of making this dependent on DataSources was to be able to have one line in your podfile if you wanted to use swift: ` pod 'Vokoder/Swift'`.  Changing this means you'll need two if you want to use the datasources.  If cocoapods is smart enough to combine multiple subspecs into a single framework and including multiple subspecs doesn't cause headaches, then I'm ok with this.

designatednerd: @chillpop When I do this in a certain application:

screen shot 2017-03-18 at 9 54 29 am

It seems to work fine: 

screen shot 2017-03-18 at 9 54 39 am

designatednerd: fwiw, i've also figured out that this works even if you add different subspecs to different targets.

PR #134, opened 2018-09-25
swift 4.2

khaptonstall:
## Description:

Made the following updates for Swift 4.2 support:

`.travis.yml`:
- Updated to use Xcode10.0
- Added tests for iOS 12

`Vokoder.podspec`
- Updated swift version to 4.2
- Updated podspec version to 7.0.0

`Paging Data Source Example`, `SampleProject`, and `SwiftSampleProject`
- Ran `pod install` to update Vokoder version 
- Update Cocoapods version to latest stable (1.5.3)
- Updated any Swift 4.2 syntax changes


khaptonstall:
Looking into why `testDeletingObjectsOnTempContextGetsSavedToMainContext` times out on iOS 12. Each test will pass if triggered individually. Also, it looks to be related to the test that runs before it, `testContextChain`, because if I comment it out or make it run after the delete test, the test suite runs just fine.

bryanluby:
Thanks @khaptonstall

bryanluby:
Tagged and pushed to trunk

PR #135, opened 2019-05-17
feature/swift5

khaptonstall:
## Description:

Made the following updates for Swift 5.0 support:

`.travis.yml`:
- Updated to use Xcode10.2
- Added tests for iOS 12.2

`Vokoder.podspec`
- Updated swift version to 5.0
- Updated podspec version to 8.0.0

`Paging Data Source Example`, `SampleProject`, and `SwiftSampleProject`
- Ran `pod install` to update Vokoder version 
- Update Cocoapods version to latest stable (1.6.2)
- Update Fastlane version to latest stable (2.123.0)
- Updated any Swift 5.0 syntax changes

Also removed "deprecated" methods in `VOKCoreDataManager+Swift.swift`. Quoting "deprecated" as they were marked `@available(*, deprecated: 4.1.0, message: "use managedObject(ofClass:inContext:) instead")` where `4.1.0` was meant to signify the podpsec version, but the `@available` attribute knows nothing about the podspec version, so the compiler actually interpreted that as "Deprecated on all platforms of version 4.1.0". Now, in Swift 5, that throws a compiler warning since no platform is specified. I just removed the methods altogether since they've been "deprecated" for multiple version releases now, but if it would be preferred to leave them in, I'd either need to make the deprecation based on a Swift language version or for specific versions of each platform (iOS, macOS, tvOS)


bryanluby (approved):
Thanks for this Swift 5 update @khaptonstall

bryanluby:
Tagged and pushed to trunk

